PR #28, opened 2019-08-01
initial Spark support

mfiess:


src/platforms/browser/WebPlatform.mjs
```diff
@@ -112,6 +113,109 @@ export default class WebPlatform {
         return cancelCb;
     }
 
+    createRoundRect(cb, stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor) {
+        if (fill === undefined) fill = true;
+        if (strokeWidth === undefined) strokeWidth = 0;
+
+        let canvas = this.getDrawingCanvas();
```
g-zachar: Prefer const over let

src/platforms/spark/TextTextureRendererSpark.mjs
```diff
@@ -0,0 +1,22 @@
+import TextTextureRenderer from "../../textures/TextTextureRenderer.mjs";
+
+export default class TextTextureRendererSpark extends TextTextureRenderer {
+    _draw() {
+        let textTextureRender = this;
+        let sparkText = sparkscene.create({ t: "text", text:textTextureRender._settings.text, pixelSize:textTextureRender._settings.fontSize*textTextureRender.getPrecision()});
+
+        let drawPromise = new Promise(function(resolve, reject) {
```
g-zachar: Style: could just return new Promise
Style: prefer arrow function expression

src/textures/TextTexture.mjs
```diff
@@ -428,7 +429,12 @@ export default class TextTexture extends Texture {
 
         return function(cb) {
             const canvas = this.stage.platform.getDrawingCanvas();
-            const renderer = new TextTextureRenderer(this.stage, canvas, args);
+            let renderer;
```
g-zachar: Style:
```javascript
const renderer = Utils.isWeb ?
  new TextTextureRenderer(this.stage, canvas, args) : 
  new TextTextureRendererSpark(this.stage, canvas, args);
```

src/tools/Tools.mjs
```diff
@@ -15,134 +14,30 @@ export default class Tools {
...
         let factory = (cb, stage) => {
-            cb(null, this.createRoundRect(stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor));
-        }
+            stage.platform.createRoundRect(cb, stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor);
+        };
         let id = 'rect' + [w, h, strokeWidth, strokeColor, fill ? 1 : 0, fillColor].concat(radius).join(",");
         return Tools.getCanvasTexture(factory, id);
     }
 
-    static createRoundRect(stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor) {
```
g-zachar: This function exposed to framework's public API and cannot be removed or changed freely. If you want to change implementation, you need change original function to be adapter, e.g.

```javascript
static createRoundRect(stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor) {
    stage.platform.createRoundRect(null, stage, w, h, radius, strokeWidth, strogeColor, fill, fillColor);
}
```

src/tools/Tools.mjs
```diff
@@ -15,134 +14,30 @@ export default class Tools {
...
 
         let factory = (cb, stage) => {
-            cb(null, this.createShadowRect(stage, w, h, radius, blur, margin));
+            stage.platform.createShadowRect(cb, stage, w, h, radius, blur, margin);
         };
         let id = 'shadow' + [w, h, blur, margin].concat(radius).join(",");
         return Tools.getCanvasTexture(factory, id);
     }
 
-    static createShadowRect(stage, w, h, radius, blur, margin) {
```
g-zachar: Ditto

src/tools/Tools.mjs
```diff
@@ -15,134 +14,30 @@ export default class Tools {
...
         let factory = (cb, stage) => {
-            this.createSvg(cb, stage, url, w, h);
-        }
+            stage.platform.createSvg(cb, stage, url, w, h);
+        };
         let id = 'svg' + [w, h, url].join(",");
         return Tools.getCanvasTexture(factory, id);
     }
-
-    static createSvg(cb, stage, url, w, h) {
```
g-zachar: Ditto

src/platforms/browser/WebPlatform.mjs
```diff
@@ -112,6 +113,109 @@ export default class WebPlatform {
...
+        ctx.lineTo(x + w, y + h - radius[2]);
+        ctx.arcTo(x + w, y + h, x + w - radius[2], y + h, radius[2]);
+        ctx.lineTo(x + radius[3], y + h);
+        ctx.arcTo(x, y + h, x, y + h - radius[3], radius[3]);
+        ctx.lineTo(x, y + radius[0]);
+        ctx.arcTo(x, y, x + radius[0], y, radius[0]);
+        ctx.closePath();
+
+        if (fill) {
+            if (Utils.isNumber(fillColor)) {
```
g-zachar: Missing Utils namespace

src/platforms/browser/WebPlatform.mjs
mfiess: Moved this code back to Tools.mjs based on feedback.  This code is now left as is (and unmodified) in Tools.mjs.

src/platforms/spark/TextTextureRendererSpark.mjs
mfiess: Now using the array syntax.

src/textures/TextTexture.mjs
mfiess: Updated based on suggestion.

src/tools/Tools.mjs
mfiess: Updated the code based on your feedback and suggestion.

src/tools/Tools.mjs
mfiess: Updated the code based on your feedback and suggestion.

src/platforms/browser/WebPlatform.mjs
mfiess: Updated the code based on your feedback and suggestion.

src/tools/Tools.mjs
```diff
@@ -121,8 +129,12 @@ export default class Tools {
 
     static getSvgTexture(url, w, h) {
         let factory = (cb, stage) => {
-            this.createSvg(cb, stage, url, w, h);
-        }
+            if (Utils.isWeb) {
```
g-zachar: Is this necessary? I can see only one implementation of createSvg, the one in platform is gone, isn't it?

src/tools/Tools.mjs
```diff
@@ -75,7 +79,11 @@ export default class Tools {
         }
 
         let factory = (cb, stage) => {
-            cb(null, this.createShadowRect(stage, w, h, radius, blur, margin));
+            if (Utils.isWeb) {
```
g-zachar: I don't see createShadowRect in platform namespace anymore.

src/tools/Tools.mjs
```diff
@@ -15,8 +15,12 @@ export default class Tools {
         }
 
         let factory = (cb, stage) => {
-            cb(null, this.createRoundRect(stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor));
-        }
+            if (Utils.isWeb) {
+                cb(null, this.createRoundRect(stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor));
+            } else {
+                stage.platform.createRoundRect(cb, stage, w, h, radius, strokeWidth, strokeColor, fill, fillColor);
```
g-zachar: I don't see implementation of createRoundRect in platform namespace anymore.

src/textures/TextTexture.mjs
```diff
@@ -428,7 +429,9 @@ export default class TextTexture extends Texture {
 
         return function(cb) {
             const canvas = this.stage.platform.getDrawingCanvas();
-            const renderer = new TextTextureRenderer(this.stage, canvas, args);
+            const renderer = Utils.isWeb ?
```
g-zachar: Is there a way to detect spark platform specifically? E.g. by checking sparkscene object in global namespace, but maybe there is a better way?

src/tools/Tools.mjs
mfiess: stage.platform.createSvg is now in the Lightning-Spark repo platform area.

src/tools/Tools.mjs
mfiess: stage.platform.createShadowRect is now in the Lightning-Spark repo platform area.

src/tools/Tools.mjs
mfiess: stage.platform.createRoundRect is now in the Lightning-Spark repo platform area.

src/textures/TextTexture.mjs
mfiess: Sure, i would be happy to add a way to detect the Spark platform specifically.  Does something like the following look okay to you?

`
(this.stage.platform.platformName && typeof this.stage.platform.platformName === "function" && this.stage.platform.platformName() === "spark")
`

If not, can you please suggest a check that would be best?  Also, do you have a preference if I should encapsulate this check somewhere?

src/tools/Tools.mjs
g-zachar: I'm not sure if I understood you correctly, but I don't see any files commited other that TextTextureRenderer.js in spark platform area.

src/textures/TextTexture.mjs
g-zachar: More like, how can application detect if it runs on spark platform? Is there e.g. global object avaliable only on spark, or perhaps there is spark specific userAgent?
Current implementation (although very crude) gives something like this:

```javascript
Utils.isNode = (typeof window === "undefined");
Utils.isWeb = (typeof window !== "undefined");
Utils.isWPE = Utils.isWeb && (navigator.userAgent.indexOf("WPE") !== -1)
```

Some of those probably need to be altered to account for spark aswell.

src/tools/Tools.mjs
mfiess: Currently, the files are located here: https://github.com/pxscene/Lightning-Spark

It's a separate repo similar to Lightning-node.

src/textures/TextTexture.mjs
mfiess: Currently, Spark doesn't have a navigator object (similar to node).  We can use the same check as we have for node `(typeof window === "undefined")`, but do you have have a suggestion how to best differentiate between node and Spark?

mfiess: Or (if you don't have a preference), I can add something to the platform area to differentiate between node and Spark.  Let me know and I will update.

g-zachar: Can sparkscene object be used to determine platform? I didn't see it imported anywhere so I assume it's global object specific to spark environment?

mfiess: Yeah, great idea!  I'll add that as part of the check.

mfiess: Added a Utils.isSpark with the check suggested above.

src/textures/TextTexture.mjs
```diff
@@ -428,7 +429,9 @@ export default class TextTexture extends Texture {
 
         return function(cb) {
             const canvas = this.stage.platform.getDrawingCanvas();
-            const renderer = new TextTextureRenderer(this.stage, canvas, args);
+            const renderer = Utils.isSpark ?
+                new TextTextureRendererSpark(this.stage, canvas, args) :
```
g-zachar: Since spark platform implementation code is in Lightning-Spark, wouldn't it make more sense to move TextTextureRendererSpark there aswell?

mfiess: I added the file the the platform area in src/platforms/spark because it depends TextTextureRenderer.  Is this not the best place to put it (given its dependency on Lightning's TextTextureRenderer)?  If you feel that the best spot is to move from the platforms area of Lightning into the Lightning-Spark repo then I will do it.  Please let me know.

g-zachar: I see. Ideal solution would be for lightning to be platform agnostic (or just for web) and have forks of Lightning with different implementations (node or spark) OR have all platform support inside upstream lightning. From what I've seen, we currently have somewhat inbetween not very elegant solution so let me think a little bit about best solution here.

g-zachar: Ok, I think it's best to have TextTextureRendererSpark and copy of TextTextureRenderer in Lightning-Spark.

mfiess: Moving TextTextureRendererSpark and a copy of TextTextureRenderer out of Lightning and into Lightning-Spark repository became pretty complex because TextTextureRenderer also had a dependency on StageUtils and Utils and bringing them over as well didn't seem like the best option.  Also, making an update to TextTextureRenderer will then require an update to the same class in the Lightning-Spark repo.

To make things simpler, I removed the TextTextureRendererSpark class and added a platform.drawText method the Lightning-Spark repo.  This should keep the changes small and clean while also achieving your requirement of having all the Spark specific code inside its own repository.  Please review and let me know if any additional changes are needed.  Thanks!

PR #35, opened 2019-08-21
Added text overflow option

### LINKED ISSUE #26: How can have letter wrapping for text
For the text component, 

we have the option of wrapping it by providing wordWrapWidth. But if a single long word is given then, it becomes empty

Eg:
> **Bigsampleword**

Expected
>**Bigsam...**

Actual
> **Empty**

**It would much help if we have an option to wrap with letters rather than omitting the complete word**

g-zachar:
Resolves #26

### DIFF
--- a/src/textures/TextTexture.mjs
+++ b/src/textures/TextTexture.mjs
@@ -97,6 +97,17 @@ export default class TextTexture extends Texture {
         }
     }
 
+    get textOverflow() {
+        return this._textOverflow;
+    }
+
+    set textOverflow(v) {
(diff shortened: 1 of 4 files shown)


g-zachar:
How to use text overflow:
```javascript
text: {
  text: 'Hello world',
  wordWrap: false, // turn off newline word wrapping
  wordWrapWidth: 300, // width where text block must fit
  textOverflow: 'ellipsis', // possible values are: 'ellipsis', 'clip', any custom string
}
```

Ill update documentation as soon as possible.

PR #51, opened 2019-09-19
fix(fireAncestors): prevent method from being called on component tha…

erikhaandrikman:
…t emits

### DIFF
--- a/src/application/Component.mjs
+++ b/src/application/Component.mjs
@@ -446,7 +446,7 @@ export default class Component extends Element {
         }
 
         const parent = this._getParentSignalHandler();
-        if(parent){
+        if (parent) {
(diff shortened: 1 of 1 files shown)


src/application/Component.mjs
```diff
@@ -445,7 +445,10 @@ export default class Component extends Element {
             throw new Error("Ancestor event name must be prefixed by dollar sign.");
         }
 
-        return this._doFireAncestors(name, args);
+        const parent = this._getParentSignalHandler();
+        if(parent){
```
g-zachar: style: missing space after if and before bracket.

erikhaandrikman: this now has been changed

PR #54, opened 2019-09-26
fix(Transition/shaders): remove circular dependency

### LINKED ISSUE #52: Warnings Emitted from Rollup When Building
When I rollup the project - the folliong circular dependencies are called out - 

```
(!) Circular dependency: node_modules/wpe-lightning/src/tree/Element.mjs -> node_modules/wpe-lightning/src/animation/Transition.mjs -> node_modules/wpe-lightning/src/tree/Element.mjs
(!) Circular dependency: node_modules/wpe-lightning/src/renderer/webgl/shaders/GrayscaleShader.mjs -> node_modules/wpe-lightning/src/renderer/c2d/shaders/GrayscaleShader.mjs -> node_modules/wpe-lightning/src/renderer/webgl/shaders/GrayscaleShader.mjs
```

Would it be possible to get these cleaned up?

erikhaandrikman:
Resolve #52


src/animation/Transition.mjs
```diff
@@ -197,6 +204,9 @@ export default class Transition extends EventEmitter {
     }
 
 }
+Transition.getGetterProxy = null;
```
g-zachar: What do we need those proxy methods for?

erikhaandrikman: Well actually we don't need them, i've committed a change

PR #61, opened 2019-10-17
Removed hard test dependencies

g-zachar:
Those libraries should be downloaded from registry rather than included in this repository.


woutermeek:
LGTM

PR #75, opened 2019-11-29
Factor update method

g-zachar:
Makes webkit's JIT compiler have easier time optimising this function.


woutermeek:
@g-zachar LGTM, do you think need to run specific tests or can add this as is?

g-zachar:
@woutermeek If browser team confirms that it indeed solves JIT compiler issue or at least improves on it then we can merge.

PR #97, opened 2020-01-15
Feature/es5 bundle

michielvandergeest:


devtools/lightning-inspect.es5.js
```diff
@@ -0,0 +1,825 @@
+(function (global, factory) {
```
g-zachar: Build results should not be version controlled. Instead, I can update npm package with es5 bundles.

michielvandergeest: @g-zachar the dist folder (and automatic build step upon each commit with husky) is added intentionally to make sure the built versions are available even if you npm install straight from GitHub

g-zachar: That's true, but regular developer should use npm registry package in the first place. Install from github should be only in special cases e.g. in your case, where you are actively contributing, but in this case having to rebuild bundles each time shouldn't be that much of a problem.

wouterlucas:
@g-zachar and @michielvandergeest lets park the dist vs npm discussion, I'll schedule something to talk about that including @woutermeek and @erikhaandrikman (and anyone that wants to chime in).

For now let's focus on ES5 support as that's quite high prio and touches a few components more. This is one of the first.

PR #108, opened 2020-02-04
additional Spark updates

mfiess:


erikhaandrikman:
I think it would be wise to create a separate Spark renderer, so we can remove the conditional statements and let the Spark provide it's own DefaultShader. I would suggest to take a look at our [canvas2d renderer](https://github.com/WebPlatformForEmbedded/Lightning/blob/master/src/renderer/c2d/C2dRenderer.mjs), use it as a starting point and add an extra rendering mode to the [Stage](https://github.com/WebPlatformForEmbedded/Lightning/blob/10105cb33ac0b4be63c8ab36e157c318a99856b6/src/tree/Stage.mjs#L66) 

This would also prevent future Spark render logic in our WebGL renderer

PR #120, opened 2020-03-02
Changed target to chrome 39.

michielvandergeest:


michielvandergeest:
Hi! Reasons for this change are:

- aligning with targets used in SDK / CLI
- impact is minimal (only difference is the typeof as you said, which is IE only anyway)
- being specific about what we (want to) support. Currently the lowest _official_ target is chrome v39 (used to be chrome 44). Using IE 11 as a target, suggests we “officially support” it

g-zachar:
Fair enough, although personally I see significant difference between officially supporting platform and setting it as target for bundling. Some of platforms that we may want officially support in the future (especially embedded ones) are not included in babel presets so eventually I'd expect we would end up setting lowest possible preset (`ie 11` in this case) and adding custom polyfills.

PR #216, opened 2020-08-13
Added support for base64 images.

g-zachar:
Allows to use base64 encoded string as image source, e.g.

```javascript
Img : {
  src: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA...',
  w: 200,
  h: 200
}
```

### DIFF
--- a/src/platforms/browser/ImageWorker.mjs
+++ b/src/platforms/browser/ImageWorker.mjs
@@ -144,7 +144,7 @@ const createWorker = function() {
     }
 
     ImageWorkerServer.isPathAbsolute = function(path) {
-        return /^(?:\/|[a-z]+:\/\/)/.test(path);
+        return /^(?:\/|[a-z]+:\/\/)/.test(path) || path.substr(0, 5) == 'data:';
     };
 
     ImageWorkerServer.prototype._receiveMessage = function(e) {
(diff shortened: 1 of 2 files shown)


src/platforms/browser/ImageWorker.mjs
```diff
@@ -144,7 +144,7 @@ const createWorker = function() {
     }
 
     ImageWorkerServer.isPathAbsolute = function(path) {
-        return /^(?:\/|[a-z]+:\/\/)/.test(path);
+        return /^(?:\/|[a-z]+:\/\/)/.test(path) || path.substr(0, 5) == 'data:';
```
michielvandergeest, line 147: any reason to not test for strict equality?

g-zachar: Using loose comparison operator for strings is a historical convention rather than anything else, but there are some obscure examples why you would actually want to avoid strict comparison between string types ([here](https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons))

```javascript
"abc" == new String("abc")    // true
"abc" === new String("abc")   // false
```

erikhaandrikman (approved):

Issue #226, opened 2020-09-01
Setting `crossOrigin: "Anonymous"` attribute on Images causes CORS issues on some platforms


Moussa:
We're testing a Lightning application on the PS4 and have run into an issue where images fail to load from some domains. Our image endpoints responses include the required `Access-Control-Allow-Origin: *` header for CORS but for some reason the PS4 browser still fails to load the images. We do not see this issue when running in any other browser, so we do not believe it's any configuration issue. It's just the PS4 browser security behaviour is unusual and unfortunately we don't have control over this behaviour.

We have verified that removing the `crossOrigin` attribute on images resolves the issue on the PS4 and so we're proposing that we add a stage option to disable adding this attribute for these problematic platforms. I will PR the changes.

g-zachar:
Fixed with #231.

PR #227, opened 2020-09-01
Add option for setting crossOrigin attribute on images (#226)

Moussa:
This PR adds a stage option to Lightning for setting the `crossOrigin: "Anonymous"` attribute on images. This allows problematic platforms to disable the attribute and avoid CORS issues.

### DIFF
--- a/src/platforms/browser/WebPlatform.mjs
+++ b/src/platforms/browser/WebPlatform.mjs
@@ -100,8 +100,9 @@ export default class WebPlatform {
                 image.cancel();
             }
         } else {
+            const setAnonymousCrossOrigin = this.stage.getOption('setAnonymousCrossOrigin');
             let image = new Image();
-            if (!(src.substr(0,5) == "data:")) {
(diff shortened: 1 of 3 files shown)


g-zachar:
Hi @Moussa,

since this is more of a platform specific fix rather than an useful option I wonder if we can identify PS4 platform during runtime, similar to this:
```javascript
Utils.isWeb = (typeof window !== "undefined") && (typeof sparkscene === "undefined");
Utils.isWPE = Utils.isWeb && (navigator.userAgent.indexOf("WPE") !== -1);
Utils.isSpark = (typeof sparkscene !== "undefined");
Utils.isNode = (typeof window === "undefined") || Utils.isSpark;
```
and make proper fix for PS4 specifically.

Best regards

Moussa:
@g-zachar That sounds reasonable, the PS4 is the only platform we've run into this issue with. I'll close this PR and re-open with the new approach 👍

PR #255, opened 2020-11-06
Feature/shader example

jfboeve:
- Added FadeOut shader.
- Added Spinner2 shader.
- Updated RoundRectangle.
- Updated RadialGradient.
- Updated Vignette.
- Changed examples folder structure.
- Added examples for regularly used shaders.


src/renderer/webgl/shaders/LoaderShader.js
```diff
@@ -0,0 +1,156 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import DefaultShader from "./DefaultShader.mjs";
+import StageUtils from "../../../tree/StageUtils.mjs";
+
+export default class LoaderShader extends DefaultShader {
```
g-zachar, line 23: Could be renamed to SpinnerShader2 to indicate this is new and improved version, the old one can would remain for backwards compatibility.

examples/shaders/fade-out/fade-out.html
```diff
@@ -0,0 +1,152 @@
...
+<body>
+<script type="module">
+    import lng from '../../../src/lightning.mjs';
+
+    class FadeOutExample extends lng.Application {
+        static _template() {
+            return {
+                w: 1920, h: 1080,
+                FadeOut1: {
+                    x: 90,
```
g-zachar, line 40: Examples could be improved using `flex` and variables like `const SRC='...'; const SIZE = 300`. Also, it's better to use local images rather than external ones.

g-zachar (changes requested):

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -21,115 +21,128 @@ import DefaultShader from "./DefaultShader.mjs";
...
 
-    set y(v) {
-        this._y = v;
+    set innerColor(argb) {
+        this._ic = argb;
+        this._normalizedIC = this._getNormalizedColor(argb);
         this.redraw();
     }
 
-    set radiusX(v) {
```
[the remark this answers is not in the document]
jfboeve: Fixed, the radiusY is now supported again. Do however notice that it will be slightly different, this is because the old shader never projected the correct diameters. Somewhere the assumption was made that diameters === radius. Which is not.

src/renderer/webgl/shaders/FadeOutShader.mjs
```diff
@@ -0,0 +1,166 @@
...
+        this.redraw();
+    }
+
+    get left() {
+        return this._fade[3];
+    }
+
+    set fade(v) {
+        if(typeof v === 'string') {
+            v = v.replace('px', '').split(' ').map(num => +num);
```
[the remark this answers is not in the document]
jfboeve: Agreed, I should have stripped it sooner, as it was a bit of a useless feature.

src/renderer/webgl/shaders/RoundedRectangleShader.mjs
```diff
@@ -25,14 +25,31 @@ export default class RoundedRectangleShader extends DefaultShader {
...
+        this._fillColor = this._getNormalizedColor(0x00ffffff);
         this._strokeColor = this._getNormalizedColor(0x00ffffff);
     }
 
     set radius(v) {
-        if (Array.isArray(v)) {
-            this._radius = v;
-        } else {
+        if(typeof v === 'string') {
+            v = v.split(' ').map(num => parseInt(num, 10));
```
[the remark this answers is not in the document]
jfboeve: Ditto

src/renderer/webgl/shaders/RoundedRectangleShader.mjs
```diff
@@ -156,7 +226,8 @@ RoundedRectangleShader.fragmentShaderSource = `
         }
 
         float b = boxDist(vTextureCoord.xy * resolution - halfRes, halfRes - 0.005, r);
-        vec4 tex = mix(vec4(0.0), color * fillColor, fillMask(b));
-        gl_FragColor = mix(tex, vColor * strokeColor, innerBorderMask(b, stroke));
+        vec4 color = mix(texture2D(uSampler, vTextureCoord) * vColor, fillColor * alpha, fill);
```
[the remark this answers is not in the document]
jfboeve: I reverted some of it back, but i think the strokeColor wasn't functional if the texture color/vColor unless it was white so that's why i started comparing it to an alpha color. The background/texture itself should blend like it did before. I think we need to have a convo about this soon.

src/renderer/webgl/shaders/LoaderShader.js
jfboeve: Ok changed it to Spinner2

examples/shaders/fade-out/fade-out.html
jfboeve: added the const for SRC and SIZE. But won't incorporate flex as I don't want to overcomplicate the examples with that functionality(hit me up in chat for more info).

src/renderer/webgl/shaders/LoaderShader.js
michielvandergeest: Would SpinnerShader2 be a public name? I.e. as an app developer you would use it in your code as `Lightning.shaders.SpinnerShader2` or something like that?

g-zachar: Something like that, you would use `Lightning.shaders.Spinner2`.

src/lightning.mjs
```diff
@@ -98,6 +100,8 @@ const lightning = {
         Pixelate: PixelateShader,
         RadialFilter: RadialFilterShader,
         RoundedRectangle: RoundedRectangleShader,
+        Loader: LoaderShader,
+        FadeOut: FadeOutShader,
         Hole: HoleShader,
         Vignette: VignetteShader,
         Spinner: SpinnerShader,
```
g-zachar, line 107: missing Spinner2 in export

src/renderer/webgl/shaders/LoaderShader.js
michielvandergeest: My 2 cents:

if possible we should just replace the old Spinner shader with the new version (Spinner2) if that one is better. I imagine it's possible to add some support for the old API (if there are any changes) to not break backward compatibility.

the only reason I'd keep 2 spinners side by side, is when Spinner2 isn't capable of doing something that Spinner1 could do. But in that case I'd choose a more descriptive name for the new shader than just appending '2'

g-zachar: The main difference is that the old shader can be used on top of existing texture as overlay (all its shortcomings come from this feature), while new one doesn't have it but instead is overall better quality. Also, any changes that have impact on visuals of the existing apps I would consider backwards incompatible and this shader falls into that category. Spinner2 is not such a bad name, especially because of recently released lightning 2.0, since it could either mean new and improved version but also as a reference to new lightning version.

michielvandergeest: I think the goal should be to have only one Spinner Shader. And if we need more, I’d really prefer not to call it Spinner2 if we can prevent that. 

My understanding so far is that “Spinner1” requires a background texture and “Spinner2” doesn’t (but also doesn’t display it when specified), right @jfboeve ?

So wouldn’t we achieve BC by just adding the functionality to display a background texture to the new Spinner?

src/renderer/webgl/shaders/SpinnerShader.mjs
```diff
@@ -60,12 +60,12 @@ export default class SpinnerShader extends DefaultShader {
     }
 
     set color(v) {
-        this._color = v;
+        this._color = new Float32Array(StageUtils.getRgbaComponentsNormalized(v));
```
g-zachar, line 63: Those changes break the shader completely as values initialised in constructor for `_color` and `_backgroundColor` are not converted. I prefer those to be reverted as I can't see any reason for them.

src/renderer/webgl/shaders/SpinnerShader.mjs
```diff
@@ -60,12 +60,12 @@ export default class SpinnerShader extends DefaultShader {
...
         this.redraw();
     }
 
     set backgroundColor(v) {
-        this._backgroundColor = v;
+        this._backgroundColor = new Float32Array(StageUtils.getRgbaComponentsNormalized(v));
```
g-zachar, line 68: Breaking change.

src/renderer/webgl/shaders/SpinnerShader.mjs
```diff
@@ -81,13 +81,12 @@ export default class SpinnerShader extends DefaultShader {
         this._setUniform('period', this._period, this.gl.uniform1f);
         this._setUniform('angle', this._angle, this.gl.uniform1f);
         this._setUniform('smooth', this._smooth, this.gl.uniform1f);
-        this._setUniform('color', new Float32Array(StageUtils.getRgbaComponentsNormalized(this._color)), this.gl.uniform4fv)
-        this._setUniform('backgroundColor', new Float32Array(StageUtils.getRgbaComponentsNormalized(this._backgroundColor)), this.gl.uniform4fv)
+        this._setUniform('color', this._color, this.gl.uniform4fv);
```
g-zachar, line 84: Breaking change.

src/renderer/webgl/shaders/SpinnerShader.mjs
```diff
@@ -81,13 +81,12 @@ export default class SpinnerShader extends DefaultShader {
...
+        this._setUniform('backgroundColor', this._backgroundColor, this.gl.uniform4fv);
```
g-zachar, line 85: Breaking change.

g-zachar:
Spinner and Spinner2 are still incompatible:

```javascript
    static _template() {
        const SIZE = 250;
        const STYLE = {rect: true, w: SIZE, h: SIZE}
        const URL = 'https://image.shutterstock.com/image-vector/example-stamp-260nw-426673501.jpg';
        const S1 = {radius: 50, width: 35, color: 0xffff5544}
        const S2 = {src: URL}
        const S3 = {period: 2, backgroundColor: 0x11ffffff, color: 0xcc000000}
        return {
            w: SIZE * 4,
            flex: {direction: 'row', wrap: true},
            Rect1: {...STYLE, color: 0xff443322, shader: {type: lng.shaders.Spinner, ...S1}},
            Rect2: {...STYLE, color: 0xff996677, shader: {type: lng.shaders.Spinner}, ...S2},
            Rect3: {...STYLE, color: 0xff55cc11, shader: {type: lng.shaders.Spinner, ...S3}},
            Rect4: {...STYLE, color: 0xff665577, shader: {type: lng.shaders.Spinner}},

            Rect5: {...STYLE, color: 0xff443322, shader: {type: lng.shaders.Spinner2, ...S1}},
            Rect6: {...STYLE, color: 0xff996677, shader: {type: lng.shaders.Spinner2}, ...S2},
            Rect7: {...STYLE, color: 0xff55cc11, shader: {type: lng.shaders.Spinner2, ...S3}},
            Rect8: {...STYLE, color: 0xff665577, shader: {type: lng.shaders.Spinner2}},
        }
    }
```

src/renderer/webgl/shaders/SpinnerShader.mjs
jfboeve: reverted

src/renderer/webgl/shaders/SpinnerShader.mjs
jfboeve: reverted

src/renderer/webgl/shaders/SpinnerShader.mjs
jfboeve: reverted

src/renderer/webgl/shaders/SpinnerShader.mjs
jfboeve: reverted

src/renderer/webgl/shaders/OutlineShader.mjs
```diff
@@ -45,7 +45,7 @@ export default class OutlineShader extends DefaultShader {
             col[1] = col[1] * col[3];
             col[2] = col[2] * col[3];
 
-            this._color = col;
+            this._color = new Float32Array(col);
```
g-zachar, line 48: Same as with `SpinnerShader`, this is a breaking change.

src/renderer/webgl/shaders/OutlineShader.mjs
```diff
@@ -60,7 +60,7 @@ export default class OutlineShader extends DefaultShader {
     setupUniforms(operation) {
         super.setupUniforms(operation);
         let gl = this.gl;
-        this._setUniform("color", new Float32Array(this._color), gl.uniform4fv);
+        this._setUniform("color", this._color, gl.uniform4fv);
```
g-zachar, line 63: Ditto

src/renderer/webgl/shaders/SpinnerShader.mjs
```diff
@@ -60,12 +60,12 @@ export default class SpinnerShader extends DefaultShader {
     }
 
     set color(v) {
-        this._color = v;
+        this._color = v//vStageUtils.getRgbaComponentsNormalized(v);
```
g-zachar, line 63: Style

src/renderer/webgl/shaders/SpinnerShader.mjs
```diff
@@ -60,12 +60,12 @@ export default class SpinnerShader extends DefaultShader {
...
         this.redraw();
     }
 
     set backgroundColor(v) {
-        this._backgroundColor = v;
+        this._backgroundColor = v//StageUtils.getRgbaComponentsNormalized(v);
```
g-zachar, line 68: Style

g-zachar (changes requested):
Also please remove `.DS_Store` file leftovers.

src/renderer/webgl/shaders/SpinnerShader2.mjs
```diff
@@ -0,0 +1,184 @@
...
+
+    set stroke(value) {
+        this._stroke = Math.abs(value);
+    }
+
+    get stroke() {
+        return this._stroke;
+    }
+
+    set time(ts) {
```
g-zachar, line 51: This shouldn't be part of API, it's very hard to make good use of it. It should be either removed or replaced by `angle` to make initial angle configuration possible.

jfboeve: No because it is not related to the angle

jfboeve: That setter is so you can synchronize your spinners rotation with other spinners

g-zachar: Wouldn't it be better for it to have constant time baseline set as default so that all spinners are in sync by default?

src/renderer/webgl/shaders/LoaderShader.js
g-zachar: I can understand in the future more spinners could be added e.g. https://projects.lukehaas.me/css-loaders/ . The main visual difference between old and new one is that new one has circular tip. @michielvandergeest, perhaps you can suggest naming scheme. Also, trying to make the two compatible turned out to be not feasible so we're stuck with both (which is fine because by default they look different)

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -63,73 +49,116 @@ export default class RadialGradientShader extends DefaultShader {
...
+        }
+
+        if(this._radius === 0) {
+            this._radius = owner.w * 0.5;
+        }
+
+        this._setUniform('innerColor', this._normalizedIC, this.gl.uniform4fv);
+        this._setUniform('fill', StageUtils.getRgbaComponentsNormalized(this._oc)[3], this.gl.uniform1f);
+        this._setUniform('outerColor', this._normalizedOC, this.gl.uniform4fv);
+        this._setUniform('pivot', new Float32Array(this._anchor),  this.gl.uniform2fv);
```
g-zachar, line 155: Should be `_pivot`

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -139,16 +168,27 @@ RadialGradientShader.fragmentShaderSource = `
...
     #endif
+    
+    #define PI 3.14159265359
+    
     varying vec2 vTextureCoord;
     varying vec4 vColor;
-    varying vec2 pos;
     uniform sampler2D uSampler;
+    uniform vec2 resolution;
+    uniform vec2 anchor;
```
g-zachar, line 178: Should be `pivot`

src/renderer/webgl/shaders/OutlineShader.mjs
g-zachar: Not a proper revert. Please note that on some of the platforms, passing regular array causes problems in shader (https://github.com/rdkcentral/Lightning/pull/235/files)

src/renderer/webgl/shaders/RadialGradientShader.mjs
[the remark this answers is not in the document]
g-zachar: I did some tests and it looks to me the original `RadialGradient` shader is completely broken so I don't think it's used anywhere (although might be wrong), so even though technically it breaks compatibility it actually fixes the shader so we'll keep the change, worst case scenario we'll revert it.

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -63,73 +49,116 @@ export default class RadialGradientShader extends DefaultShader {
...
+        }
+        if(this._y) {
+            this._pivot[1] = this._y / owner.h;
+        }
+
+        if(this._radius === 0) {
+            this._radius = owner.w * 0.5;
+        }
+
+        this._setUniform('innerColor', this._normalizedIC, this.gl.uniform4fv);
```
g-zachar, line 152: Please note that on some of the platforms it is required to convert arrays to `Float32Array` for shader to work.

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -63,73 +49,116 @@ export default class RadialGradientShader extends DefaultShader {
...
+        this._setUniform('fill', StageUtils.getRgbaComponentsNormalized(this._oc)[3], this.gl.uniform1f);
```
g-zachar, line 153: Ditto.

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -63,73 +49,116 @@ export default class RadialGradientShader extends DefaultShader {
...
+        this._setUniform('outerColor', this._normalizedOC, this.gl.uniform4fv);
```
g-zachar, line 154: DItto.

g-zachar:
Still some `.DS_Store` files lingering around.

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -21,36 +21,22 @@ import DefaultShader from "./DefaultShader.mjs";
...
+        this._normalizedIC = this._getNormalizedColor(this._ic);
+        this._oc = 0x00ffffff;
+        this._normalizedOC = this._getNormalizedColor(this._oc);
+        this._radius = 0;
     }
 
     set radiusX(v) {
-        this._radiusX = v;
-        this.redraw();
+        this.radius = v;
```
g-zachar, line 35: Shouldn't it be `_radius`?

src/renderer/webgl/shaders/RadialGradientShader.mjs
```diff
@@ -21,36 +21,22 @@ import DefaultShader from "./DefaultShader.mjs";
...
-
-    set y(v) {
-        this._y = v;
-        this.redraw();
+        this._pivot = [0, 0];
+        this._ic = 0xffffffff;
+        this._normalizedIC = this._getNormalizedColor(this._ic);
+        this._oc = 0x00ffffff;
+        this._normalizedOC = this._getNormalizedColor(this._oc);
+        this._radius = 0;
```
g-zachar, line 31: `_radiusY` is not initialised

g-zachar (approved):

jfboeve (commented):
Don't know what happened here..

PR #264, opened 2020-12-01
Component.mjs: accept functions as signals

pluzecki:


examples/signals/web-bundle/signals.html
```diff
@@ -0,0 +1,84 @@
...
+                        Bg: {
+                            src: "../../landscape.jpg", scale: 1,
+                        },
+                        Box: {
+                            type: Box,
+                            x: w => w / 2,
+                            y: h => h / 2,
+                            mount: 0.5,
+                            signals: {
+                                signalHandler1: ctx.signalHandler1,
```
g-zachar, line 63: `true` and `string` handlers could be added to example (see [docs](https://rdkcentral.github.io/Lightning/docs/components/communication/signal#docsNav))

g-zachar (changes requested):

g-zachar (approved):

examples/signals/signals.html
```diff
@@ -0,0 +1,120 @@
...
+                        Bg: {
+                            src: "../landscape.jpg", scale: 1,
+                        },
+                        Box: {
+                            type: Box,
+                            x: w => w / 2,
+                            y: h => h / 2,
+                            mount: 0.5,
+                            signals: {
+                                signalHandler1: ctx.signalHandler1,
```
victor-wm, line 87: This introduces a bug.

As the template is `static`, it will be evaluated once you construct the first component and its value will be saved into the class. That parsed template will have closure over the context being passed to `_template(ctx)` meaning that `ctx.signalHandler1` will belong to the first component instance, while new components created from the same class will not have the right context as `_template(` will not be executed again.

I would advise against keeping this in the codebase until it's figured out how to give the proper instance to the template.

PR #270, opened 2020-12-07
Feature: isfocused property both in object and in DOM when inspector and debug is true

jhderojasUVa:
**Problem:**

There's no way to know on inspector if a component has focus or not.

**Solution:**

An attribute is created called `isfocused` that tells you if it has focus. Also, is a property of the object that you can get. Also the old FOCUS message on the console was on when you set debug as true on the settings.json file.

### DIFF
--- a/devtools/lightning-inspect.js
+++ b/devtools/lightning-inspect.js
@@ -748,11 +748,9 @@ window.attachInspector = function({Element, ElementCore, Stage, Component, Eleme
     Object.defineProperty(ElementCore.prototype, 'isfocused', {
       // We are working on the core!
       get: function() {
-        console.log('Inspector geeting is focused')
         return this._isfocused
       },
       set: function(v) {
-        console.log('Inspector setting is focused')
         this._isfocused = v
 
         // Put the attribute 'isfocused' as the current value
--- a/devtools/lightning-inspect.js
+++ b/devtools/lightning-inspect.js
@@ -27,7 +27,7 @@ window.attachInspector = function({Element, ElementCore, Stage, Component, Eleme
     window.mutationCounter = 0;
     window.mutatingChildren = false;
     var observer = new MutationObserver(function(mutations) {
-        var fa = ["x", "y", "w", "h", "alpha", "mountX", "mountY", "pivotX", "pivotY", "scaleX", "scaleY", "rotation", "visible", "clipping", "rect", "colorUl", "colorUr", "colorBl", "colorBr", "color", "borderWidthLeft", "borderWidthRight", "borderWidthTop", "borderWidthBottom", "borderWidth", "borderColorLeft", "borderColorRight", "borderColorTop", "borderColorBottom", "borderColor", "zIndex", "forceZIndexContext", "renderToTexture", "renderToTextureLazy", "renderOffscreen", "colorizeResultTexture", "texture"];
+        var fa = ["x", "y", "w", "h", "alpha", "mountX", "mountY", "pivotX", "pivotY", "scaleX", "scaleY", "rotation", "visible", "clipping", "rect", "colorUl", "colorUr", "colorBl", "colorBr", "color", "borderWidthLeft", "borderWidthRight", "borderWidthTop", "borderWidthBottom", "borderWidth", "borderColorLeft", "borderColorRight", "borderColorTop", "borderColorBottom", "borderColor", "zIndex", "forceZIndexContext", "renderToTexture", "renderToTextureLazy", "renderOffscreen", "colorizeResultTexture", "texture", "isfocused"];
         var fac = fa.map(function(v) {return v.toLowerCase()});
 
         mutations.forEach(function(mutation) {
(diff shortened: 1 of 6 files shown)


jhderojasUVa:
I'm not very confident with this and I like to be double and triple checked!

devtools/lightning-inspect.js
```diff
@@ -740,6 +743,21 @@ window.attachInspector = function({Element, ElementCore, Stage, Component, Eleme
         }
     });
 
+    // defining the isfocused property and attribute
+    ElementCore.prototype._isfocused = false;
```
michielvandergeest, line 747: Maybe we should camelCase `isFocused` for readability and to follow the pattern of the other properties?

jhderojasUVa: You are right!. Done!

src/tree/Element.mjs
```diff
@@ -1596,6 +1596,19 @@ export default class Element {
         }
     }
 
+    /**
+     * isfocused property of the Elemenet
+     *
+     * This property sets and gets if the element has focus
+     */
+    get isfocused() {
```
newmanja, line 1604: This may shadow the existing hasFocus() and hasFinalFocus on _Component_; is there a way to lean on that existing feature (can a non-Component have focus?? sorry I'm new here :) ).

newmanja: MR with proposed refactor for that raised with jhderojasUVa directly.

g-zachar: I second this, there shouldn't be any need to pollute the core with additional flags like `isfocused`.

g-zachar (changes requested):

src/application/Application.mjs
```diff
@@ -102,12 +102,31 @@ export default class Application extends Component {
...
 
         if (!Application.booting && notOverridden) {
             this.updateFocusSettings();
         }
+
+        if (this.__options.debug) {
+            const newFocusedComponent = this.__focusedComponent();
+
+            // Testing for updateDebugFocusState safeties against lightning inspector not having been attached.
+            prevFocusedComponent && prevFocusedComponent.updateDebugFocusState && prevFocusedComponent.updateDebugFocusState();
```
g-zachar, line 123: This dependency injection style of programming is out of ordinary for Lightning's codebase. Instead, you could monkey patch `__updateFocus` method to follow existing code's standards.

Another problem: inspector shouldn't be tied to `debug` flag - it should only influence the log, while inspector's `focus` attribute should operate regardless of the flag.

src/application/Application.mjs
```diff
@@ -102,12 +102,31 @@ export default class Application extends Component {
...
+            newFocusedComponent && newFocusedComponent.updateDebugFocusState && newFocusedComponent.updateDebugFocusState();
+
+            if (prevFocusedComponent !== newFocusedComponent) {
+                console.log('FOCUS ' + (newFocusedComponent && newFocusedComponent.getLocationString()));
```
g-zachar, line 127: Please add `[Lightning]` tag to the log. This distinction is helpful when there are other logging mechanisms used on top.

devtools/lightning-inspect.js
```diff
@@ -740,6 +743,10 @@ window.attachInspector = function({Element, ElementCore, Stage, Component, Eleme
         }
     });
 
+    Element.prototype.updateDebugFocusState = function() {
+        val(this, 'hasfinalfocus', this.hasFinalFocus(), false);
```
g-zachar, line 747: Shouldn't it be `isfocused`?

newmanja: Using `hasFinalFocus` seemed to better match the way that developers would understand the focus; being able to see the value in the inspector helps the QAs on our project, and having the two teams be aligned helps us.

At some point we might also look to expose `hasFocus` (which is true for the whole chain of focus-containing elements) but that's more complex to implement, and not immediately needed.

The description for this pull request does still mention 'isfocused' but imho that's out of date.

devtools/lightning-inspect.js
```diff
@@ -116,6 +116,9 @@ window.attachInspector = function({Element, ElementCore, Stage, Component, Eleme
                                 case "colorizeResultTexture":
                                     pv = false
                                     break;
+                                case "isfocused":
```
newmanja, line 119: I don't think we need (want) this `isfocused` nor the one in `var fa =` above.

src/application/Application.mjs
newmanja: @jhderojasUVa if this console log doesn't shadow one that was in the original code, maybe we can just remove this if clause?

devtools/lightning-inspect.js
g-zachar: I would assume developers care much more for final element rather than whole chain which is an implementation detail that `hasFinalFocus` indirectly exposes and may lead to confusion. Focus chain could be exposed with more descriptive flag indicating chain implementation in the future if ever necessary.

g-zachar:
Follow up: #304

PR #282, opened 2021-02-22
Add sideEffects false flag for Webpack

chiefcll:
Help webpack know that Lightning can be tree shaked - https://webpack.js.org/configuration/optimization/#optimizationsideeffects

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "name": "@lightningjs/core",
   "version": "2.2.0",
   "license": "apache",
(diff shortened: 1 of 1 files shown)


g-zachar:
I don't think adding this flag would help webpack tree shake in any way because of how lightning module is currently built (it exports single object). However, you raised a valid concern that perhaps we should redesign it so that tree shaking of unused shaders (as there are more and more) and components (which should be deprecated and eventually removed anyway) is possible.

PR #292, opened 2021-03-16
Allow binding objects to text property

### LINKED ISSUE #283: bindProp is not working with the `text` shorthand
Hey all, when using `bindProp()` to return a bound `text` texture property one would expect it to work with this syntax:

```
public static _template() {
  const boundText = this.bindProp('text', (ctx) => {
    return { text: ctx.text, color: ctx.color }
  });
  
  return {
    SomeComponent: {
      text: boundText
    }
  }
}
```

Lightning seems to ignore the property completely, probably due to the way it parses shorthand properties for `text`.

Any thoughts?

g-zachar:
Resolves #283

### DIFF
--- a/src/application/Component.mjs
+++ b/src/application/Component.mjs
@@ -172,7 +172,13 @@ export default class Component extends Element {
                 if (key === "text") {
                     const propKey = cursor + "__text";
                     loc.push(`var ${propKey} = ${cursor}.enableTextTexture()`);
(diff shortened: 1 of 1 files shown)


victor-wm (approved):
LGTM. Thanks for this @g-zachar

erikhaandrikman (approved):

PR #297, opened 2021-04-09
Allow textures to be permanent (not cleared by GC)

g-zachar:


scthunderbolt:

Issue #68, opened 2019-11-07
Support for mouse/magic remote


kirandcunha:
Hi Team,
Is there a support for mouse/magic remote available here? I did not find any reference in the documentation.

erikhaandrikman:
There is currently no support for mouse / magic remote. Magic remote support is on our roadmap

markus101:
I've taken a look at the branch that starts support for this (https://github.com/WebPlatformForEmbedded/Lightning/tree/touch-support), but it is missing handling for the `enableTouch` option as it's never added to the app options via `_setOptions`. 

Once I fixed the options locally I tested on webOS and found touch support is not sufficient for magic remote support as the magic remote behaves like a regular mouse. Once things were changed to use `mousedown` instead of `touchstart` it worked as expected on webOS.

erikhaandrikman:
Thanks for reporting, will close the ticket for now and take this in account

g-zachar:
Support was added in 2.4.0.

PR #300, opened 2021-05-06
Pointer support

harryfb:
Adds hover, click & scroll functionality (intended for use with the LG Magic Remote but applicable to other pointers as it uses standard pointer events). Based on the POC described in #68.

To determine the hover/click target, the pointer is localised inside active candidate components. The item with the the highest z-index is assumed the target (id when z-indices are the same). The component tree is searched upwards for an assigned handler. Note, as the hover implementation makes use of the *mousemove* event, the hover handler is only called when the target component changes, removing excessive handler calls.

Both top-down and bottom-up handlers have been implemented for handling scroll wheel events as required.

### DIFF
--- a/src/application/Application.mjs
+++ b/src/application/Application.mjs
@@ -54,7 +54,7 @@ export default class Application extends Component {
             });
         }
 
-        if (this.getOption("enableClicks")) {
+        if (this.getOption("enablePointer")) {
             this.stage.platform.registerClickHandler((e) => {
                 this._receiveClick(e);
             });
@@ -88,7 +88,7 @@ export default class Application extends Component {
             8: "Back",
             27: "Exit"
         });
-        opt('enableClicks', false);
+        opt('enablePointer', false);
     }
 
     __construct() {
@@ -426,38 +426,19 @@ export default class Application extends Component {
     fireTopDownClickHandler(obj) {
         const {clientX, clientY} = obj;
         let children = this.stage.application.children;
+
         // reverse so while loops searches top down
         let affected = this._findChildren([], children).reverse();
-        let n = affected.length;
-        const withinClickableRange = [];
-
-        // loop through affected children
-        // and perform collision detection
-        while (n--) {
-            const child = affected[n];
-            const precision = this.stage.getRenderPrecision();
-            const ctx = child.core.renderContext;
-
-            const cx = ctx.px * precision;
-            const cy = ctx.py * precision;
-            const cw = child.finalW * ctx.ta * precision;
-            const ch = child.finalH * ctx.td * precision;
-
-            if (cx > this.stage.w || cy > this.stage.h) {
-                continue;
-            }
-            if (this._testCollision(clientX, clientY, cx, cy, cw, ch)) {
-                withinClickableRange.push(child);
-            }
-        }
+        let clickableChildren = this._withinClickableRange(affected, clientX, clientY);
         
-        if (withinClickableRange.length) {
-            withinClickableRange.sort((a,b) => {
+        if (clickableChildren.length) {
+            clickableChildren.sort((a,b) => {
                 return a.id > b.id ? 1: -1;
             });
-            let n = withinClickableRange.length;
+            let n = clickableChildren.length;
+
             while (n--) {
-                const child = withinClickableRange[n];
+                const child = clickableChildren[n];
                 if (child && child["_handleClick"]) {
                     child._handleClick();
                     break;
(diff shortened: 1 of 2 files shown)


harryfb:
Examples of how to use this feature in practice:
https://github.com/harryfb/lightning-pointer-examples

src/application/Application.mjs
```diff
@@ -405,6 +421,195 @@ export default class Application extends Component {
...
+                } else if (event === "_captureScrollUp") {
+                    child._captureScrollUp();
+                    return true;;
+                }         
+            }
+        }
+        return false
+    }
+
+    fireBottomUpScrollWheelHandler(event, obj) {
```
g-zachar, line 461: Scroll handler should pass the event to the hook so that the developer can access its deltaX and deltaY values. As a consequence, there is no need for up/down distinction, just `handleScroll`.

src/application/Application.mjs
```diff
@@ -405,6 +421,195 @@ export default class Application extends Component {
...
+        const {clientX, clientY} = obj;
+
+        if (clientX <= this.stage.w && clientY <= this.stage.h) {
+            // force a drawFrame to set any missing element dimensions
+            this.stage.drawFrame();
+            this.stage.application.fireBottomUpHoverHandler(obj);
+        }
+    }
+
+    fireBottomUpHoverHandler(obj) {
```
g-zachar, line 519: It would be useful to provide `_handleUnhover` hook as well, e.g:

```javascript
    fireBottomUpHoverHandler(obj) {
        const {clientX, clientY} = obj;
        const target = this._getTargetChild(clientX, clientY);

        // Only fire handler when pointer target changes
        if (target !== this.__hoveredChild) {
            if (this.__hoveredChild) {
                let child = this.__hoveredChild;
                while (child !== null) {
                    if (child && child["_handleUnhover"]) {
                        child._handleUnhover(target);
                        break;
                    }
                    child = child.parent;
                }
            }

            let child = target;
            this.__hoveredChild = target;

            while (child !== null) {
                if (child && child["_handleHover"]) {
                    child._handleHover(target);
                    break;
                }
                child = child.parent;
            }
        }
    }
```

src/application/Application.mjs
```diff
@@ -405,6 +421,195 @@ export default class Application extends Component {
...
+        return false
+    }
+
+    _receiveClick(e) {
+        const obj = e;
+        const {clientX, clientY} = obj;
+
+        if (clientX <= this.stage.w && clientY <= this.stage.h) {
+            // force a drawFrame to set any missing element dimensions
+            this.stage.drawFrame();
```
g-zachar, line 488: Frame redraw is not necessary.

src/application/Application.mjs
```diff
@@ -405,6 +421,195 @@ export default class Application extends Component {
...
+        }
+    }
+
+    _receiveHover(e) {
+        const obj = e;
+        const {clientX, clientY} = obj;
+
+        if (clientX <= this.stage.w && clientY <= this.stage.h) {
+            // force a drawFrame to set any missing element dimensions
+            this.stage.drawFrame();
```
g-zachar, line 514: Frame redraw is not necessary.

harryfb:
Thank you for reviewing. I have finished making those changes - let me know if you have any other modifications.

src/application/Application.mjs
```diff
@@ -405,6 +421,187 @@ export default class Application extends Component {
...
+            this.stage.application.fireBottomUpHoverHandler(obj);
+        }
+    }
+
+    fireBottomUpHoverHandler(obj) {
+        const {clientX, clientY} = obj;
+        const target = this._getTargetChild(clientX, clientY);
+
+        // Only fire handlers when pointer target changes
+        if (target && (target !== this.__hoveredChild)) {
```
g-zachar, line 504: `_handleUnhover` should fire even when new target is `null`.

harryfb: Good point, I've removed this condition.

src/application/Application.mjs
```diff
@@ -405,6 +421,187 @@ export default class Application extends Component {
         return;
     }
 
+    _recieveScrollWheel(e) {
+        const obj = e;
+        const {clientX, clientY,} = obj;
```
g-zachar, line 426: Style (coma)

src/application/Application.mjs
```diff
@@ -405,6 +421,187 @@ export default class Application extends Component {
...
+
+        // Search tree bottom up for a handler
+        while (child !== null) {
+            if (child && child[event]) {
+                child._handleScroll(obj);
+                return true;
+            }
+            child = child.parent;
+        }
+        return false
```
g-zachar, line 463: Style (semicolon)

src/application/Application.mjs
```diff
@@ -405,6 +421,187 @@ export default class Application extends Component {
...
+            } else {
+                return a.id > b.id ? 1: -1;
+            }
+        });
+
+        if (hoverableChildren.length) {
+            // Assume target has highest zIndex (id when zIndex equal)
+            return hoverableChildren.slice(-1)[0];
+        } else {
+            return null
```
g-zachar, line 550: Style (semicolon)

g-zachar (changes requested):

g-zachar (approved):

PR #302, opened 2021-06-07
Better pointer support

g-zachar:
- fixed collision detection on rtt and clipping settings
- added collision flag for customisation
- added examples

### DIFF
--- a/examples/mouse-pointer/basic-usage.html
+++ b/examples/mouse-pointer/basic-usage.html
@@ -2,7 +2,7 @@
  If not stated otherwise in this file or this component's LICENSE file the
  following copyright and licenses apply:
 
- Copyright 2020 RDK Management
+ Copyright 2020 Metrological
 
  Licensed under the Apache License, Version 2.0 (the License);
  you may not use this file except in compliance with the License.
--- a/examples/mouse-pointer/basic-usage.html
+++ b/examples/mouse-pointer/basic-usage.html
@@ -39,6 +39,7 @@
                         h: 200,
                         rect: true,
                         color: 0xffaa7777,
+                        collision: true,
                         Text: {text: {text: 'Click me!'}, x: 25, y: 50, rotation: 0.5}
                     }
                 }
--- a/examples/mouse-pointer/basic-usage.html
+++ b/examples/mouse-pointer/basic-usage.html
@@ -1,0 +1,84 @@
+<!--
+ If not stated otherwise in this file or this component's LICENSE file the
+ following copyright and licenses apply:
+
+ Copyright 2020 RDK Management
+
+ Licensed under the Apache License, Version 2.0 (the License);
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8" />
+    <script src="../../../devtools/lightning-inspect.js"></script>
+</head>
+<body style="margin: 0; padding: 0">
+    <script type="module">
+        import lng from '../../../src/lightning.mjs';
+        //attachInspector(lng)
+
+        window.onload = function() {
+
+            const ENABLE_POINTER = true;
+
+            class Card extends lng.Component {
+                static _template() {
+                    return {
+                        w: 200,
+                        h: 200,
+                        rect: true,
+                        color: 0xffaa7777,
+                        Text: {text: {text: 'Click me!'}, x: 25, y: 50, rotation: 0.5}
+                    }
+                }
+
+                _handleClick() {
+                    this.animation({
+                        duration: 0.2,
+                        actions: [
+                            {p: 'scale', v: {0: 1.0, 0.5: 1.1, 1: 1.0}}
+                        ]
+                    }).start();
+                }
+
+                _handleHover() {
+                    this.setSmooth('color', 0xff77aa77);
+                }
+
+                _handleUnhover() {
+                    this.setSmooth('color', 0xffaa7777);
+                }
+
+
+            }
+            class BasicUsageExample extends lng.Application {
+                static _template() {
+                    return {
+                        Card1: {type: Card, x: 50, y: 50},
+                        Card2: {type: Card, x: 50 + 200 + 25, y: 50},
+                        Card3: {type: Card, x: 50 + 400 + 50, y: 50, collision: false, alpha: 0.5}
+                    }
+                }
+
+            }
+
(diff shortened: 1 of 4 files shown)


src/tree/Element.mjs
```diff
@@ -112,6 +112,12 @@ export default class Element {
         this._w = 0;
 
         this._h = 0;
+
+        /**
+         * When enablePointer setting is on, this flag will be checked if the element should respond
+         * to the mouse events.
+         */
+        this._collision = true;
```
sky-pedrofernandes, line 120: Apologies for barging in on this, but semantically, wouldn't it make more sense for the flag to represent that a certain element was "transparent", collision-wise?

Also, creating a large number of elements is something that is already costly on some very resource-limited devices - do we really need to initialise another flag for something that will probably (and this is me assuming stuff) be the exception, or not used at all in projects that don't need to support the LG Wand?

g-zachar: Hi @sky-pedrofernandes

Can you elaborate on "transparency" a little bit more?

While it's true that the `_collision` flag could be hidden behind the setting option flag for idiomatic reasons, the memory footprint it leaves is negligible not only on its own but also compared to all other memory overhead present to even take into consideration. When that becomes necessary, proper specific optimizations for RAM usage can be applied instead that will have a significant impact.

sky-pedrofernandes: I've been wrapping my head around the best way to explain my point of view and apologise upfront should I fail miserably.

The way it stands, right now, with the flag having the default value of `collision: true` what we're saying is that **all elements** will be eligible for collision-detection unless you specify otherwise, correct? To put it another way, what's being provided to the Lightning developer is the ability to say which elements *should not be taken into consideration*.
In this situation what we'll effectively be doing is then specifying which elements should be `transparent` during collision-testing, so perhaps the flag could reflect that explicitly. As I mentioned above, it's a simple semantic nuance and something that, personally, made sense to me.

Now, thinking about this a bit more, if we analyse it exclusively from a performance point of view, I would argue that the default value should actually be `collision: false`, so that you would really only collision-test elements you **knew were going to allow interactions with the pointer**.
I do understand that this would make it a bit harder to use as you'd need to explicitly mark the elements you wanted to consider for pointer interaction, but it would avoid a lot of collision checks.
In this case, I think the `collision` flag would actually be much more appropriate than `transparency`.

Finally, regarding the memory footprint - I'm not saying that this would be the biggest of our problems, and you can always argue (correctly, I might add) that the effect will be mostly negligible. But it won't be 0. I do apologise for being particularly picky with this, but some of the devices we're working with are really limited in terms of performance and anything that could be made to avoid impacting that even further, would be most welcome - even more so because those devices won't actually support any pointer devices.

g-zachar: @sky-pedrofernandes

That's a very good point, indeed having all the elements eligible for collision by default is not ideal (although they are initially narrowed down to all active elements). I reversed the logic but as a consequence added a branch marking mechanism to handle nested cases as with current implementation this would otherwise put stress on the developer to mark the entire branch and not just the final element manually. I might revisit the entire implementation to find a more elegant solution.

examples/mouse-pointer/basic-usage.html
```diff
@@ -0,0 +1,85 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8" />
+    <script src="../../../devtools/lightning-inspect.js"></script>
```
erikhaandrikman, line 24: Path to devtools directory is incorrect

examples/mouse-pointer/basic-usage.html
```diff
@@ -0,0 +1,85 @@
...
+</head>
+<body style="margin: 0; padding: 0">
+    <script type="module">
+        import lng from '../../../src/lightning.mjs';
```
erikhaandrikman, line 28: Incorrect path

examples/mouse-pointer/scroll.html
```diff
@@ -0,0 +1,114 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8" />
+    <script src="../../../devtools/lightning-inspect.js"></script>
```
erikhaandrikman, line 24: Incorrect path

examples/mouse-pointer/scroll.html
```diff
@@ -0,0 +1,114 @@
...
+    <style>
+        body.disable-scroll {
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+</head>
+<body style="margin: 0; padding: 0">
+    <script type="module">
+        import lng from '../../../src/lightning.mjs';
```
erikhaandrikman, line 34: Incorrect path

examples/mouse-pointer/scroll.html
```diff
@@ -0,0 +1,114 @@
...
+                _handleUnhover(target) {
+                    this.setSmooth('color', 0x11cccccc)
+                    if (target.__ref == 'BackToTop') {
+                        target.setSmooth('color', 0xffcccccc);
+                        target.setSmooth('scale', 1.0);
+                    }
+                }
+
+                _handleClick(target) {
+                    if (target.__ref == 'BackToTop') {
```
erikhaandrikman, line 88: `target.ref`

examples/mouse-pointer/scroll.html
```diff
@@ -0,0 +1,114 @@
...
+                    this.setSmooth('color', 0x22ccffcc);
+                    if (target.__ref == 'BackToTop') {
+                        target.setSmooth('color', 0xffffffff);
+                        target.setSmooth('scale', 1.2);
+                    }
+                }
+
+                _handleUnhover(target) {
+                    this.setSmooth('color', 0x11cccccc)
+                    if (target.__ref == 'BackToTop') {
```
erikhaandrikman, line 81: `target.ref`

src/tree/Element.mjs
```diff
@@ -1411,6 +1412,24 @@ export default class Element {
...
+        this._collision = v;
+    }
+
+    _updateCollision() {
+        if (this.collision && this.__parent && this.__parent.collision === undefined) {
+            /**
+             * Mark collision as 2 to indicate atleast one descendant has collision enabled.
+             * This narrows down the search for elements with active collision.
+            */
+            this.__parent.collision = 2;
```
erikhaandrikman, line 1429: Any reasoning behind marking it as `2` Other then: https://github.com/rdkcentral/Lightning/blob/6dfd288945b6207f55391262db86e66e85381d77/src/application/Application.mjs#L564 evaluating to true?

erikhaandrikman (changes requested):

src/tree/Element.mjs
g-zachar: Currently, it allows to filter out all the unnecessary elements early on before further evaluation in response to mouse events (if the elements were evaluated to `true` instead they would be needlessly processed). Other than that, this distinction might become useful later should this feature need any alterations.

erikhaandrikman (approved):

Issue #323, opened 2021-08-25
Mouse click event exclude the element that has clipping:true property


Balavaradh:
When the component contains clipping: true and contains _handleClick event, it was skipped from affected child elements inside the _withinClickableRange

![image](https://user-images.githubusercontent.com/30177859/130732897-c2f594c8-471b-4667-bedc-b5579fa074e9.png)

Either removing the scissor condition or add a property to exclude this condition will solve this issue.

g-zachar:
Hi @Balavaradh,

can you elaborate on this issue a bit more, perhaps provide an example?

Best regards

PR #325, opened 2021-08-25
Mouse event should exclude the scissor check that has clipping: true property

balaexpo:
Issue detail: #323


Balavaradh:
I have accepted the licensing on the CLA page, but why it is still showing me an error CLA is not signed yet

image

scthunderbolt:

PR #355, opened 2021-12-01
Inspector, update root div when canvas width update

Marine-Dos-Wiztivi:
I observed that width and height of the root div for the inspector is retrieved based on a timeout of 1s when inspect scripts is loaded. If the width and / or height of the canvas is modified after this delay, the root div of inspector keep a width and height of 0. As a result, when using Chrome inspector, we do not see the overlay of inspected elements (ie blue rectangle).

Regards

### DIFF
--- a/devtools/lightning-inspect.js
+++ b/devtools/lightning-inspect.js
@@ -193,15 +193,26 @@ window.attachInspector = function({Application, Element, ElementCore, Stage, Com
             var root = document.createElement('DIV');
             document.body.appendChild(root);
             var self = this;
-            setTimeout(function() {
-                var bcr = self.stage.getCanvas().getBoundingClientRect();
+            let updateRootStyleFromCanvas = function (bcr) {
                 root.style.left = bcr.left + 'px';
                 root.style.top = bcr.top + 'px';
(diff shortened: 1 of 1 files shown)


g-zachar (approved):

g-zachar:
Looks like a likely race condition for larger apps, thanks for contributing.

evgeniy-legkun:
@g-zachar @Marine-Dos-Wiztivi On some devices that do not support ResizeObserver API (like Ziggo EOS) it crashes the app with the error "Unhandled Promise Rejection: ReferenceError: Can't find variable: ResizeObserver".
Please, use more robust conditions, do not try to call ResizeObserver directly, but through the environment global object (window, self, global), window.ResizeObserver. See the polyfill as an example https://www.npmjs.com/package/resize-observer-polyfill

g-zachar:
Hi @evgeniy-legkun,

That's unfortunate. Can you check if `window.ResizeObserver != null` will mitigate the issue for you? I will not be adding any polyfills to lightning as it will cascade into code bloat pretty quickly and will negatively impact bundle size and performance. I'd rather revert the change if it has negative side effects or mitigate it and rework the fix as to not use ResizeObserver.

Best regards

PR #378, opened 2022-04-20
Improve support for vertical alignment on advanced text renderer.

itsjamie:
Using metadata from fonts with the advanced text renderer users will now be able to set their font baseline ratio.

This enables users to vertically position text accurately within the rasterized text canvas, so that further positioning is accurate.

In addition, the calculation of the height of the text canvas has changed when line height is greater than the font size and the previous calculation of maxLines is not met. 

This is one section I'd appreciate a double-check on. I'm not super familiar with rendering a baseline that is `'ideographic'`. If the font was parsed and the default line height was set from metrics in the fonts, I think it would be ok (the change), but given that the line height defaults to font size when unset, it's possible this change could result in some fonts rendering slightly off-canvas if no baseline ratio is given.

The calculations as they existed are an average that approximates many fonts, but for fonts with high baselines, you end up with text that tends to be higher than intended in the canvas representation when a middle vertical alignment is given.
As well, the canvas was always larger than other renderers would attribute to the same text, which was an issue for content-based fits, like Flexbox.

Visuals:

With a proper baseline, line height, and vertical alignment set to middle, here is the result on the rasterized canvas.

Screen Shot 2022-04-20 at 1 49 53 PM

With no font baseline information, but line height and vertical alignment set to middle here is the result.

Screen Shot 2022-04-20 at 1 47 26 PM

Original Visual:

If you remove the change to the canvas size, and have no font baseline information, here is the current result today.

image

So, I hope these visuals makes it clear. Essentially, by oversizing the canvas, you ended up closely approximating center the text for _most_ font baselines. But you have a much larger canvas than necessary, which means size of items are larger than expected.

References:
https://opentype.js.org/font-inspector.html - useful for calculating your font baseline ratio
https://docs.microsoft.com/en-us/typography/opentype/spec/os2#stypoascender
https://docs.microsoft.com/en-us/typography/opentype/spec/os2#stypodescender

### DIFF
--- a/src/textures/TextTextureRendererAdvanced.mjs
+++ b/src/textures/TextTextureRendererAdvanced.mjs
@@ -119,6 +119,8 @@ export default class TextTextureRendererAdvanced {
         renderInfo.text = this._settings.text;
         renderInfo.precision = precision;
         renderInfo.fontSize = fontSize;
+        // 0 is used as the baseline ratio if none is provided so it has no impact over the browser default rendering.
+        renderInfo.fontBaselineRatio = this._settings.fontBaselineRatio || 0.0 
         renderInfo.lineHeight = lineHeight;
         renderInfo.letterSpacing = letterSpacing;
         renderInfo.textAlign = this._settings.textAlign;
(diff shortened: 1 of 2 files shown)


src/textures/TextTextureRendererAdvanced.mjs
```diff
@@ -202,12 +204,36 @@ export default class TextTextureRendererAdvanced {
...
+            renderInfo.h = renderInfo.lineNum * renderInfo.lineHeight + fontSize / 2;
+        }
+
+        // This calculates the baseline offset in pixels from the font size.
+        // To retrieve this ratio, you would do this calculation:
+        //     (FontUnitsPerEm − hhea.Ascender − hhea.Descender) / (2 × FontUnitsPerEm)
+        //
+        // This give you the ratio for the baseline, which is then used to figure out 
+        // where the baseline is relative to the bottom of the text bounding box.
+        const baselineOffsetInPx = renderInfo.fontBaselineRatio * renderInfo.fontSize
```
itsjamie, line 227: I took care here to make sure if no value was given here that the render will not have changed. Since a zero here will result in a zero overall change to the calculations for both `top` and `middle` vertical alignment.

src/textures/TextTextureRendererAdvanced.mjs
```diff
@@ -202,12 +204,36 @@ export default class TextTextureRendererAdvanced {
...
+        if (this._settings.h) {
+            renderInfo.h = this._settings.h;
+        } else if (renderInfo.maxLines && renderInfo.maxLines < renderInfo.lineNum) {
+            renderInfo.h = renderInfo.maxLines * renderInfo.lineHeight + fontSize / 2;
+        } else if (renderInfo.lineHeight > fontSize) {
+            // When lineheight is larger than the font size we're rendering, we set the height of the canvas based on the number of lines we're rendering.
+            // This makes each "line" a containing box that is line height sized, and text is positioned inside that box.
+            //
+            // Ideographic fonts may break this model, and require additional space?
+            renderInfo.h = renderInfo.lineNum * renderInfo.lineHeight
```
itsjamie, line 216: I did change this, because this was increasing the size of elements to be larger than necessary when used via content-based fits like Flexbox.

g-zachar:
Might be useful in some cases, thanks!

g-zachar (approved):

PR #388, opened 2022-06-13
Added glReadPixels support after rendering

ricardo-a-alves-alb:
Tested on Tizen 6.0.

The main issue was that when `this._performRender();` was called, it caused flickerings and sometimes the elements even disappeared from the screen.

Since this is something you activate through a specific stage configuration, it looks good to me to merge this into a new version of the Lightning core. We really need this feature/fix because we are targeting these devices and this PR fixes it.

### DIFF
--- a/src/tree/Stage.mjs
+++ b/src/tree/Stage.mjs
@@ -186,6 +186,7 @@ export default class Stage extends EventEmitter {
         opt('canvas2d', false);
         opt('platform', null);
         opt('readPixelsBeforeDraw', false);
+        opt('readPixelsAfterDraw', false);
     }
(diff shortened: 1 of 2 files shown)


erikhaandrikman (approved):

wouterlucas:
> Could there bit a bit more context/documentation around this kind of options?
> 
> How are frameworks users supposed to find them and figure they need them for their project?

You're absolutely right. Added all 3 new flags in here: https://github.com/rdkcentral/Lightning/commit/f3daf2dfde3d46a5b15587fd74ffdd7d0c2b77ff

PR #390, opened 2022-06-20
feature(docs) Added shader documentation

jfboeve:


docs/Shaders/Hole.md
```diff
@@ -0,0 +1,73 @@
...
+
+### x
+returns a `float` with the value of the x position of the cutout.
+
+### y
+returns a `float` with the value of the y position of the cutout.
+
+### radius
+returns a `float` with the value of the radius of the corners of the cutout.
+
```
michielvandergeest, line 68: Doesn't the Holeshader also require `rtt:true` to be set? Maybe we should mention that somewhere

docs/Shaders/Perspective.md
```diff
@@ -0,0 +1,59 @@
...
+### Initialize
+
+To use the Perspective shader you need to apply it to the shader property of an element:
+
+```js
+//in template
+class MyApp extends Lightning.Application {
+    static _template() {
+        return {
+            Picture: {w: 300, h: 300, src: 'MyImage.jpg', shader: {type: Lightning.shaders.Perspective, rx: Math.PI * 0.25}}
```
michielvandergeest, line 22: Maybe we should explain a bit better what happens here with `Math.PI * 0.25`?

docs/Shaders/Spinner2.md
```diff
@@ -0,0 +1,83 @@
+# Spinner2
```
michielvandergeest, line 1: Should we mention something about Spinner (the previous version) and that it's recommended to use Spinner2?

docs/Shaders/Vignette.md
```diff
@@ -0,0 +1,100 @@
...
+this.tag('Vignette').shader = {type: Lightning.shaders.Vignette, magnitude: 3, intensity: 0.3}
+```
+
+You can customize the appearance of the Vignette shader with the following setters.
+
+
+## Setters
+
+### magnitude
+You can set the magnitude of the Vignette shader with a `float`.
```
michielvandergeest, line 37: Explain a bit better what is meant by `magnitude`?

docs/Shaders/Vignette.md
```diff
@@ -0,0 +1,100 @@
...
+
+### intensity
+You can set the intensity of the Vignette shader with a `float`.
```
michielvandergeest, line 40: Explain a bit better what is meant by `intensity`?

docs/Shaders/Hole.md
jfboeve: No it does not require rtt:true. It only requires rtt: true, if the element the shader is applied to has children and you want the children to remain as they are.

docs/Shaders/Vignette.md
jfboeve: changed to:

You can set the magnitude of the Vignette shader with a `float`. The value of magnitude determines the degrees of fading around the light source.

docs/Shaders/Perspective.md
jfboeve: added extra explanation how rotation works

docs/Shaders/Light3D.md
```diff
@@ -19,13 +19,27 @@ To use the Light 3D shader you need to apply it to the shader property of an ele
...
+Rotation in Lightning is done with radian values. You can use the following formula to calculate the correct value from degrees to radians;
+
+```
+degrees * Math.PI / 180
+```
+
+If you for example to rotate a tag 33 degrees;
+
+```
+33 * Math.PI / 180
```
michielvandergeest, line 42: This is great! :heart:

PR #391, opened 2022-06-21
Add compressed texture support

itsjamie:
Add Compressed Texture support to Lightning.

This implementation focused on the KTX1 container format to reduce the total amount of configuration required by an end user while loading images.

The WebGL renderer exposed the compressed images extensions so that at runtime a user can make the decision on what texture style to ask for.

It can contain any of the formats described and in use on the web targets. For our own testing, we utilized https://github.com/BinomialLLC/basis_universal and the UASTC style content, offline prepared to the compressed texture for the device.

This was based on a performance change with a reduction of 97% of the blocking time while loading a 1080p texture. 

Screen Shot 2022-06-20 at 1 39 50 PM

is now

Screen Shot 2022-06-20 at 1 40 24 PM

This reduction in the blocking time of a frame render makes high-performance animations with images coming on screen, a possibility.


erikhaandrikman:
This is awesome! Thanks for contributing, we will review this as soon as possible

g-zachar (approved):

PR #400, opened 2022-07-13
Enable RAF idle usage with stage option 'pauseRafLoopOnIdle'

wouterlucas:
Hi @g-zachar looking to move the RAF usage along, did some testing and it seems to work great. However to mitigate risk I'd like to make it opt-in so the default behavior remains unchanged. 

Let me know if this would work for you.


g-zachar:
LGTM

g-zachar (approved):

PR #408, opened 2022-07-29
Prevent spinner from stalling due to floating point wrap-over.

itsjamie:
Some devices, specifically saw on Samsung TVs, seem to map highp floats to be 16 bit. Leading to a frozen spinner after ~65s.

This keeps the uniform as a repeating value through the use of the modulus operator and leaves a little time beyond the period to keep the spinner looking consistent.

### DIFF
--- a/src/renderer/webgl/shaders/SpinnerShader2.mjs
+++ b/src/renderer/webgl/shaders/SpinnerShader2.mjs
@@ -20,6 +20,8 @@
 import DefaultShader from "./DefaultShader.mjs";
 import StageUtils from "../../../tree/StageUtils.mjs";
 
+// amount of time a revolution of the spinner takes.
+const REVOLUTION_TIME = 1005;
 export default class SpinnerShader2 extends DefaultShader {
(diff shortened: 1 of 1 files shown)


erikhaandrikman:
Hi, thanks for pointing this out. This will indeed fix the issue on devices where highp floats are 16 bit. Since the duration of a full spin is not a fixed value, this fix will introduce issues on spinners with a duration > 1000ms ( jumping back to the start before the full spin is completed )

Since we use the remainder of 
```
(Date.now() - SpinnerShader2.spinSync) % REVOLUTION_TIME
``` 
to determine to radius of the rotation: https://github.com/itsjamie/Lightning/blob/4515d90fbf120739e45bb48db38705ade619b528/src/renderer/webgl/shaders/SpinnerShader2.mjs#L119

we could choose to add a dynamic revolution time: 
```
this._revolutionTime = Math.min(65535, 1005 * this._period) 
```

scthunderbolt:

PR #413, opened 2022-08-26
fix: Prevent fatal exception when attempting to render 0x0 canvas texture without forceTxCanvasSource option

### LINKED ISSUE #412: Fatal exception when attempting to create 0 width/height RoundRect texture
Starting in 2.5.1 (because of [this change](https://github.com/rdkcentral/Lightning/commit/7c4b76754658ad0e1f590834b6401a619f3cda51) to the default behavior for rendering a canvas to a GL texture) if you try to create/apply a RoundRect texture with `0` stroke width and `-2` for either width or height, for example:
```js
Lightning.Tools.getRoundRect(-2, -2, 10, 0, 0xffff0000, true, 0xff00ff00)
```

You get an app crashing exception on [this getImageData() call](https://github.com/rdkcentral/Lightning/commit/7c4b76754658ad0e1f590834b6401a619f3cda51#diff-142d4910858111883a8658b068ed323f466b8139cc7fc73aed8b484660701479R80). This is because the `-2`s are converted to zeros by [createRoundRect()](https://github.com/rdkcentral/Lightning/blame/7d4dfb360142a3300bce845d10a7ab7fe9d57241/src/textures/RoundRectTexture.js#L118) and getImageData() [does not allow either width or height to be 0](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData#exceptions).

The “forceTxCanvasSource” stage option introduced in [2.7.0](https://github.com/rdkcentral/Lightning/releases/tag/2.7.0) ([PR 393](https://github.com/rdkcentral/Lightning/pull/393)) allows anyone to return to the original pre-2.5.1 behavior which does not throw an exception in this case. However an application should not need to do this.

### Expected Behavior
If "forceTxCanvasSource" is not `true`, a RoundRect created with either width/height as `-2` should simply not render anything instead of crashing the app.

frank-weindel:
Fixes #412

### DIFF
--- a/src/platforms/browser/WebPlatform.mjs
+++ b/src/platforms/browser/WebPlatform.mjs
@@ -113,8 +113,9 @@ export default class WebPlatform {
             if (Utils.isZiggo || this.stage.getOption("forceTxCanvasSource")) {
                 // Ziggo EOS and Selene have issues with getImageData implementation causing artifacts.
                 gl.texImage2D(gl.TEXTURE_2D, 0, options.internalFormat, options.format, options.type, source);
-            } else {
+            } else if (source.width > 0 && source.height > 0) {
                 // Workaround for some browsers (e.g. Tizen) as they do not convert canvas data to texture correctly, sometimes causing artifacts.
+                // Width/Height check added because of https://github.com/rdkcentral/Lightning/issues/412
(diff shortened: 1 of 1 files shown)


michielvandergeest (approved):
LGTM

PR #419, opened 2022-09-16
Fix typo on TemplateSpec.md

RobRuizR:
`TemplateSpec.md` mentions that the developer should use `ImplementsTemplateSpec`. There is a typo here, as Implements should be singular as can be seen [here](https://github.com/rdkcentral/Lightning/blob/1cd88a281a527a3d49f4d0ae082f46a406308944/src/application/Component.d.mts#L224). 

This fixes the issue.


frank-weindel:
LGTM

frank-weindel (approved):

PR #423, opened 2022-10-12
Feature/pvr container support

erikhaandrikman:


src/platforms/browser/WebPlatform.mjs
```diff
@@ -106,14 +106,15 @@ export default class WebPlatform {
     }
 
     uploadCompressedGlTexture(gl, textureSource, source, options) {
+        const view = !source.pvr ? new DataView(source.mipmaps[0]) : source.mipmaps[0];
```
michielvandergeest, line 109: Are more different types of sources besides `pvr = true` expected in the future? If so, then we could make this a bit more 'future proof'

PR #427, opened 2022-10-26
update lightning inspector root div offset - Fixes #426

### LINKED ISSUE #426: Lightning Inspector root element is not offset correctly
This PR
https://github.com/rdkcentral/Lightning/commit/477060e3e47b7dc5223a6056696ab2871255523d

Lightning components in a body tag with an offset (such as in Storybook) don't get the correct left + top offsets. This leads to all divs being off by whatever the body tag is offset by. I think this is because contentRect is not returning the true left and top offset like `getBoundingClientRect`. I think the easy fix is line 207 should be:
`updateRootStyleFromCanvas(entries[0].target.getBoundingClientRect());`
And that fixes the issue.

Current Issue: Not getting Left + Top correctly
Screen Shot 2022-10-25 at 10 40 28 PM

chiefcll:


michielvandergeest:
Looks good to me.

I guess we should consider this a hotfix (i.e. to be released outside of the quarterly release schedule)?
If so, then it's okay to do the PR straight to master. We would just need to version bump and update the changelog as well.

erikhaandrikman (approved):

PR #433, opened 2022-12-06
Add TypeScript Config Warning

frank-weindel:
A user brought to our attention that the `strictNullChecks` TS config must be set to true in order to avoid compiler errors especially when dealing with the `this.tag()` calls. See: https://github.com/rdkcentral/Lightning-SDK/issues/358#issuecomment-1339321527


michielvandergeest (approved):
LGTM

PR #436, opened 2022-12-22
Added instant freeing up of text textures to prevent memory building…

michielvandergeest:
… up when text is being changed.


wouterlucas (approved):
👀

frank-weindel (approved):

PR #437, opened 2022-12-23
Improve colorBl access on c2d/DefaultShader.mjs

jorge-graca-sky:


jfboeve (approved):

frank-weindel:
Is there a noticeable bug fixed by this change?

jorge-graca-sky:
> Is there a noticeable bug fixed by this change?
No.

michielvandergeest:
Interesting this conditional that does absolutely nothing 😉 Good find though, that the property name `colorBl` should have been prefixed with an underscore 😊

The faulty code doesn't produce any real bug, but since we are aware of this now, I suggest we do fix it one way or another

We can accept the proposed change, but that only makes sense if we expect to ever add some logic in the single color condition block. @jfboeve / @erikhaandrikman / @frank-weindel would there be a reason for that?

Otherwise I suggest we just flip the conditional around `if(vc._colorUl !== vc._colorBl)` and enter the vertical gradient logic and leave out the single color condition all together.

PR #445, opened 2023-01-23
TypeScript: Fix generic paramater Ref value getByRef() bug

### LINKED ISSUE #444: TypeScript: `getByRef()` error when using generic type param as Ref value
A user [on the forums](https://forum.lightningjs.io/t/subclassable-components-typescript/546) asked a question about how to create a subclassable base Component that takes a generic parameter that is used to inform the type of a Ref value. In diving into a potential solution I discovered that the type definitions for `getByRef()` are not allowing for a genericized Ref value type, while the definitions for `tag()` are allowing it just fine.

Example code:
```ts
import { Lightning } from '@lightningjs/sdk'

class Header extends Lightning.Component {}

export interface IPageTemplateSpec<
  T extends Lightning.Component.Constructor = Lightning.Component.Constructor,
> extends Lightning.Component.TemplateSpec {
  Header: typeof Header
  Content: T
}

export interface IPageTypeConfig extends Lightning.Component.TypeConfig {
  IsPage: true
}

export class Page<T extends Lightning.Component.Constructor = Lightning.Component.Constructor>
  extends Lightning.Component<IPageTemplateSpec<T>, IPageTypeConfig>
  implements Lightning.Component.ImplementTemplateSpec<IPageTemplateSpec<T>>
{
  static override _template(): Lightning.Component.Template<IPageTemplateSpec> {
    return {
      w: (w: number) => w,
      h: (h: number) => h,
      rect: true,
      color: 0xff0e0e0e,

      Header: {
        type: Header,
      },
      Content: undefined,
    }
  }

  protected Content = this.getByRef('Content')!
  //                                ^-------^ Error!!!
}
```

The error is:

```
Argument of type 'string' is not assignable to parameter of type 'keyof TemplateSpecRefs<IPageTemplateSpec<T>>'.
  Type '"Content"' is not assignable to type '"Header" | (TransformPossibleElement<P, IPageTemplateSpec<T>[P], never> extends never ? never : P) | ((T extends Constructor<...> ? InstanceType<...> : Element<...>) extends never ? never : "Content")'.ts(2345)
```

Example Repo: https://github.com/frank-weindel/com.domain.app.LngForum546Subclassable

frank-weindel:
Fixes #444

Implemented and tested via new and existing `tsd` tests


frank-weindel:
@uguraslan Let's hold back from merging this one. While doing some other typing work I discovered an issue with this that doesn't seem to be being detected by the tests.

frank-weindel:
Fixed the issue and added many more tests that should help solidify the whole tag/getByRef type chain. Explanation is a bit difficult to explain but here's an attempt:

Using tag() on a Component that has a lower-case property of instance type `Element` or `Component` in its TemplateSpec
resulted in tag() causing a very strange error due to "circular references". The new getByRef() implementation is based
on how tag() originally worked and saw the same exact issue.

The error was tracked back to IsTerminus<T> returning `false` for any instance type of `Element` or `Component` and the fact that
SpecToTagPaths<T> did not filter out lower-case properties. This resulted in SpecToTagPaths<T> recursively entering a lower-case
`Element`/`Component` property (which is where it should have actually terminated recursion at) and then further recursively enter
the instance's `stage`, `application`, `ctx`, etc properties ultimately resulting in a circular reference.

SpecToTagPaths<T> now only returns paths for upper case ValidRef properties. And IsTerminus<T> returns true for
`Element` and `Component`.

I've also moved many of the internal type functions that existed in Element.d.mts into internalTypes.d.mts and added comprehensive
unit tests for each of them. I've also added a test for the fixed scenario that was originally untested in components-strong-typing.test-d.ts.

This has also been tested with the TypeScript version of TMDB which is how I originally found the issue.

jfboeve (approved):

PR #452, opened 2023-02-13
TypeScript: Implement Default Loose Type Configs

frank-weindel:
Lightning's type system has the concept of Loose Template Specs ([documented here](https://lightningjs.io/docs/#/lightning-core-reference/TypeScript/Components/TemplateSpecs?id=loose-components)) which allow properties and Refs that aren't actually declared in the Template Spec to be specified as if they were `any`. By default all Components that do not specify Template Spec (i.e. use the default Template Spec) use a Loose Template Spec.

Along side the Template Spec we offer another way to make Components more type strong called the [Type Config](https://lightningjs.io/docs/#/lightning-core-reference/TypeScript/Components/TypeConfigs). Unfortunately support for a Loose Type Config was not previously considered and it was brought to my attention by the Lightning UI team that this is making it difficult to use signals and events in their Component code because due to the way their classes are structured they are unable to easily pass through a Type Config.

Here I implement a Loose Type Config in much the same way that Loose Template Specs are implemented. By default Components that do not have a specified Type Config have Loose Type Configs and you are free to use any signals and events on these components. As a bonus, one can even explicitly extend the base `TypeConfigLoose` interface if they want the best of both worlds. Any event/signal is allowed and treated (essentially) as an `any`. But explicitly defined events/signals will still be type checked.

## To-do
- [x] Implementation
- [x] Basic Unit tests
- [x] Another pass on Unit Tests
- [x] Test with TMDB project
- [x] Share build with LUI team and collect feedback
- [x] Add documentation
- [x] Wait for #445 to be merged to `dev` and then rebase this PR onto `dev`.
  - **Currently the base of this PR is the `ts-fix-getbyref` branch from this PR!!**

### DIFF
--- a/docs/TypeScript/Components/TypeConfigs.md
+++ b/docs/TypeScript/Components/TypeConfigs.md
@@ -130,3 +130,16 @@ class MyParentComponent extends Lightning.Component {
   // ...
 }
 ```
+
+## Loose Type Configs
+
+If a Type Config is not explicitly provided, a **Loose Type Config** will be used by default. This is similar to the [**Loose Template Spec**](TemplateSpecs.md). Events and Signals will be handled in a much more looser fashion, allowing you to hook up any Events/Signals regardless of if any are explicitly specified.
+
+You can also explicitly extend the base `LooseTypeConfig` interface if you want the best of both worlds. Explicitly specified Events and Signals will be type checked. But any other Event or Signal will be allowed and be able to emit/receive `any` set of parameters.
+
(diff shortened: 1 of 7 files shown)


frank-weindel:
Accidentally closed the PR

uguraslan:
Hey @frank-weindel, PR #445 has been merged into `dev` so you can do the changes regarding this PR.

wouterlucas (approved):

PR #454, opened 2023-02-16
Fix documentation and TS definitions for TextTexture fontFace

frank-weindel:
- Documentation incorrectly said the type was Integer
- TS defs were missing the string[] type


chiefcll (approved):

docs/RenderEngine/Textures/Text.md
```diff
@@ -11,7 +11,7 @@ You can use various properties to control the way in which you want to render te
 | Name | Type | Default | Description |
 |---|---|---|---|
 | `text` | String | '' | Text to display |
-| `fontFace` | Integer | null | Font family used for current text |
+| `fontFace` | String or String[] | null | Font family used for current text. If an array is provided, font familes that appear later in the array will be used as fallbacks. |
```
michielvandergeest, line 14: familes => families

also about the default, is that really `null` in practice? there is some kind of final fallback font if nothing is specified - maybe we should make some mention of that

michielvandergeest (commented):
looks good! I've added 2 comments on the documentation part

docs/RenderEngine/Textures/Text.md
frank-weindel: Yeah looking at it deeper there's a lot more detail we can add here, so I detailed it out a good bit more.

wouterlucas (approved):
LGTM

PR #458, opened 2023-02-24
Add devicePixelRatio to settings

wouterlucas:
When playing with the Lightning Touch library (https://github.com/rdkcentral/Lightning-Automotive) I ran into the problem where the devicePixelRatio was undefined.

It is needed here:
https://github.com/rdkcentral/Lightning-Automotive/blob/master/src/helpers.js#L124

And without it the touch library wont function, as the DPR will be `NaN` all collision detection will fail. By adding `devicePixelRatio` to the Lightning Stage settings the touch library will work.


uguraslan (approved):
LGTM

PR #468, opened 2023-03-27
Feature: device pixel ratio

uguraslan:
The implementation of device pixel ratio feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Changelog
 
+## v2.9.0
+*16 feb 2023*
+
+- Fixed issues related to package.json exports (#434)
+- Added a warning about `strictNullChecks` option to the TypeScript docs (#433)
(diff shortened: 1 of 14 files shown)


frank-weindel:
@erikhaandrikman This also redefines the `w` and `h` stage options. Prior to this change they represent the actual pixel size of the canvas. #471 relies on this fact. In this PR they are multiplied by the `devicePixelRatio`. For example if `devicePixelRatio` is 2, and the pixel size of the canvas as 4K (3840x2160), these stage options are transformed into 7680x4320 which I don't think is right.

jfboeve (approved):

wouterlucas (approved):
LGTM. Though we may want to document the behavior on precision if devicePixelRatio > 1 as this could lead to different behavior if people are manually adjusting the precision (from before this change) in higher DPR setups.

erikhaandrikman (approved):

PR #471, opened 2023-04-04
Fix high res text wrapping

### LINKED ISSUE #470: Text rendered at high precision levels can word-wrap incorrectly
This problem was discovered while trying to run the lng create starter app at 4K. To do this I changed the w/h stage options to 3840x2160 and set the precision to 2. Upon doing this I noticed a line break in the "Let's start building!" text line that is not there when you run the application in it's native 1920x1080.

Screenshot of lng create starter application showing the text of 'Let's start building' where the word 'building' has been incorrectly dropped to the second line

Looking at lines 110-111 in TextTextureRenderer.mjs, you can see the total maximum width of the text is by default set to a constant divided by the precision:
```ts
        // Total width.
        let width = w || (2048 / this.getPrecision());
```
This in effect makes the total possible width of text get smaller as you increase the resolution!

To fix this I will replace that value with the actual rendered pixel width of the screen:
```ts
        // Total width.
        let width = w || this._stage.getOption('w');
```

frank-weindel:
Fixes #470


wouterlucas (approved):
LGTM

Issue #450, opened 2023-02-10
TextTexture does not handle Chinese text well


Moussa:
We've observed that Chinese text within TextTextures do not line break correctly, resulting in Chinese text being rendered as a single long line that overflows the container. Perhaps Lightning should be applying https://en.wikipedia.org/wiki/Line_breaking_rules_in_East_Asian_languages? We've worked around this internally by adding a space after characters defined in https://en.wikipedia.org/wiki/Chinese_punctuation, but it would be great if Lightning could handle this internally. Other non-latin languages are likely affected.

uguraslan:
@Moussa Thank you for bringing this to our attention. To better understand the issue and reproduce it, would you be able to provide a sample app that can be created by using the [Lightning playground tool](https://lightningjs.io/playground)? It would be great if the sample app contains your work-around as well.

Moussa:
Hi @uguraslan. Sure thing, this playground app shows the issue with some sample Chinese text. If you replace the original text input with the `parseNonLatinPunctuation` parsed input you'll see improved text rendering behaviour.

[...]

PR #455, opened 2023-02-17
Integrate Vite and Vitest. Add test for TextTextureRendererUtils.

frank-weindel:
This PR is a shot at upgrading our build system to use [Vite](https://vitejs.dev/) (which uses rollup under the covers for production builds). I've attempted to replicate everything we have currently with rollup. The regular rollup build system remains untouched under the `build` npm command.

One of the benefits, and biggest motivator (for me), of switching to the Vite ecosystem is that we can take advantage of [Vitest](https://vitest.dev/) for unit testing. I've added that as well here along with a working test for the utility method created in rdkcentral/Lightning#442.

To build, simply run:
```
npm run build-vite
```
which will build all of the bundles necessary in parallel using another NPM package called [concurrently](https://www.npmjs.com/package/concurrently).

To run tests, run:
```
npm test
```

Vitest also integrates with Visual Studio Code via a plugin with the same name. And it's really nifty.

**Still to be done/worked out:**
- [x] Add lightning-inspect ES5 build
- [x] Get the copyright header plugin (rollup-plugin-license) to work
  - _Now using a custom Vite plugin which was very simple to write_
- [x] Verify that the bundles generated work as well as the pure rollup ones did.
- [x] Move this PR to the rdkcentral/Lightning (once rdkcentral/Lightning#442 is merged)
- [x] Remove vanilla rollup system (and have vite output to `dist`)
- [x] Configure TypeScript definitions to be moved to `dist` folder

PR #473, opened 2023-04-06
Implement word wrapping on zero-width breaking spaces

frank-weindel:
This PR enables word wrapping to occur on the zero-width space character `0x200B` ([Unicode Explorer](https://unicode-explorer.com/c/200B), [Wikipedia](https://en.wikipedia.org/wiki/Zero-width_space#:~:text=The%20zero%2Dwidth%20space%20character,%3B%20or%20%E2%80%8B%20.)). This enables developers to programmatically insert these characters in appropriate places for writing systems that do not use space characters to separate words without negatively affecting how the text renders when it is not word wrapped.

This feature has been implemented in both the standard TextTextureRenderer as well as TextTextureRendererAdvanced. However, I noticed that there seem to be unrelated vertical spacing issues with rendering non-Latin script like Chinese that are out of the scope of this feature.

Fixes https://github.com/rdkcentral/Lightning/issues/450 (or at least enables developers to work around the problem)

### DIFF
--- a/docs/RenderEngine/Textures/Text.md
+++ b/docs/RenderEngine/Textures/Text.md
@@ -45,9 +45,51 @@ You can use various properties to control the way in which you want to render te
 | `letterSpacing` | Integer | 0 | Letter spacing of characters |
 
 
+## Word Wrap in Non-Latin Based Languages
 
+Enabling the `wordWrap` option causes lines of text that are too long for the
+specified `wordWrapWidth` to be broken into multiple lines. Lines are broken
+only at word boundaries. In most latin script based languages (i.e. English,
+Dutch, French, etc) the space " " character is the primary separator of word
+boundaries.
 
+Many non-latin based languages (i.e. Chinese, Japanese, Thai and more) do not use spaces
+to separate words. Instead there is an assortment of rules that determine where
+word boundaries, for the purpose of line breaking, are allowed. Lightning
+currently does not implement these rules as there are many languages and writing
+systems to consider when implementing them. However, we do offer a work around
+that can be employed in your application as needed.
 
+### Zero-Width Spaces
+
+Lightning supports line breaking at [Zero-Width Space](https://en.wikipedia.org/wiki/Zero-width_space)
+(Unicode: `\u200B`) word boundaries. These characters are like normal spaces but
+take up no actual space between visible characers. You can use them in
+your text strings and Lightning will line break on them when it needs to.
+
+You may want to write a function that you funnel all of your application's
+text strings into:
+
+```js
+function addZeroWidthSpaces(text) {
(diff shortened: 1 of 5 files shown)


wouterlucas:
I guess this will only help moderated / editorial texts that the application itself already controls, in that case they could just as easily break it up manually and provide new coordinates/text elements for the next line. 

I don't see how this will help with dynamic data from API sources such as descriptions/titles/text loaded asynchronously in the application. 

Shouldn't modern browsers implement http://unicode.org/reports/tr14/#BreakOpportunities? Maybe something to figure out with the WebKit team and is there a way we could leverage the line-breaking rules built in by the browser? E.g. by providing a limiting canvas width of some kind (I know that will probably be expensive).
The more complex solution is to implement non-western line breaking rules without relying on injecting ZWSP. We could probably regex the "prohibited" characters and insert a no-break space (U+00A0) character there (as I beleve its exclusion driven right?).

frank-weindel:
> I guess this will only help moderated / editorial texts that the application itself already controls, in that case they could just as easily break it up manually and provide new coordinates/text elements for the next line.
> 
> I don't see how this will help with dynamic data from API sources such as descriptions/titles/text loaded asynchronously in the application.

I think this is a misunderstanding. All it takes to utilize these zero-width spaces is to wrap all strings in whatever function you might use to insert them just prior to handing the string off to Lightning.

So basically anytime you set `text`/`text.text` on a Lightning Element (or a Component property that is displayed). It's fairly tedious, but totally doable and solves for all strings in the application. 

We might consider making this easier for developers by providing some kind of global "displayed text transformation" hook that allows them to set up a transform function once and have it automatically applied to all text immediately before it is rendered/word wrapped.

> Shouldn't modern browsers implement http://unicode.org/reports/tr14/#BreakOpportunities? Maybe something to figure out with the WebKit team and is there a way we could leverage the line-breaking rules built in by the browser? E.g. by providing a limiting canvas width of some kind (I know that will probably be expensive).

That would seem to me to require a huge standards push. Canvas does not provide an API for drawing text with line-breaks /  word-wrapping nor does it provide an API for determining how to break lines. 

>  The more complex solution is to implement non-western line breaking rules without relying on injecting ZWSP. We could probably regex the "prohibited" characters and insert a no-break space (U+00A0) character there (as I beleve its exclusion driven right?).

I think it's quite a bit more complicated and requires us to import [unicode character line break data](https://www.unicode.org/Public/UCD/latest/ucd/LineBreak.txt) and process each character precisely according to the rules outlined. Something perhaps we want to look at doing much better for L3?

uguraslan:
Hey @frank-weindel, PR #455 has been merged into `dev`, can you please update the PR description and remove "Do not merge" label if you think everything is okay now.

wouterlucas (approved):

Issue #475, opened 2023-04-19
Incomplete regression fix on ziggo platforms


Sv-Th:
Regarding following pull: https://github.com/rdkcentral/Lightning/pull/364

A new generation of hardware was released with the following user agent:
````Mozilla/5.0 (Linux armv7l) AppleWebKit/602.1.28+ (KHTML, like Gecko) Version/9.1 Safari/601.5.17 HZN/4.46 (MN=2008C-STB-PROD;PC=EOS2STB;FV=EOS2008C-mon-web-00.05-157-ap-AL-20230209074548-na002;)````

This causes the earlier mentioned regression fix to not function for users with new generation of boxes.
This reintroduces the blocky/missing text issue of https://github.com/rdkcentral/Lightning/issues/363

The fix could be as simple as adding ``|| navigator.userAgent.indexOf('EOS2STB') !== -1`` to the isZiggo utils check. However, something to keep in mind is that if or when a new generation is introduced, the issue could resurface.

wouterlucas:
I think its better if we take the whole User Agent detection thing out and make this a switchable setting people can configure. That way it's not bound to project specific devices and it's a generic setting people can turn on. Making it future proof for next iterations of the Ziggo HW.

Sv-Th:
Apparently this generic setting has already been added (https://github.com/rdkcentral/Lightning/issues/363#issuecomment-1308359557). A similar issue has been created in the past as well. 

My mistake. I will be closing the issue and pull request in this case.

Issue #363, opened 2022-01-04
Texture rendering issue on Ziggo devices


evgeniy-legkun:
Hello everyone.
The new workaround tended to fix artifacts on some browsers (e.g. Tizen) implemented in this PR - _https://github.com/rdkcentral/Lightning/pull/345/files_ does not work correctly on Ziggo devices (EOS, Selene).
The behavior is random, most of the time we do not see the text (TextTexture texture) at all, or weird symbols are drawn.

See the screenshot 
![Screenshot 2022-01-04 at 12 17 28](https://user-images.githubusercontent.com/20604216/148044361-f97b7e1f-2a4b-4593-bf01-a0fd4fca0eea.png)

The issue comes from this line of the code that was released with version 2.5.1.
![Screenshot 2022-01-04 at 12 19 46](https://user-images.githubusercontent.com/20604216/148044599-4af9e8d4-92e6-46a8-9d33-b8a272acc6eb.png)

We had to freeze the Lightning core with the 2.5.0 version on our project and are not able to use newer versions in production because of this.
Probably this is some browser-specific issue since we reproduced it only for the devices (e.g. Ziggo) that use Gecko-based browser (Firefox in our case) underhood. Actually, I can not say what is the issue, but it is defiantly related to the getImageData() function calling. The 2d context attribute "willReadFrequently" (see https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) setting with neither "true" nor "false" did not help.

Please, investigate this issue or consider rolling back the PR mentioned above if it is possible. If need to check something on the actual device (I have Ziggo EOS) I can do it if have free time. 

Regards

g-zachar:
Hi @evgeniy-legkun,

I don't have access to Ziggo EOS. If you can find a way to distinguish your platform in JS runtime I can apply the revert for it specifically:

https://github.com/rdkcentral/Lightning/blob/669095186eb68f3d60599b498b1a0f192fdd462b/src/tree/Utils.mjs#L193-L197

Best regards

evgeniy-legkun:
Hi, @g-zachar I was looking for a way to distinguish the Ziggo platform assuming to parse a UserAgent, this does not look like a robust solution but I could not find the better one. Looks like the "lightning/sdk" Profile plugin does not have similar logic as well.

Take a look at the UserAgents examples (obtained from the devices):
1) Ziggo Selene (king of old device, has the same issue):
[...]

PR #476, opened 2023-04-19
Fix regression on new Ziggo hardware

Sv-Th:
Resolve https://github.com/rdkcentral/Lightning/issues/475


Sv-Th:
Already fixed/fixable by following setting https://github.com/rdkcentral/Lightning/issues/363#issuecomment-1308359557

PR #478, opened 2023-04-20
Fix/cleanup on destroy

wouterlucas:
Fixes a memory leak that was introduced in https://github.com/rdkcentral/Lightning/releases/tag/2.7.0 when the feature `pauseRafLoopOnIdle` (https://github.com/rdkcentral/Lightning/pull/402) is enabled.

Only happens in a scenario where you are running subsequent or multiple Lightning instances within the same browser tab. This is how Metrological often runs the Lightning applications. 

Occurs by not cleaning up the interval set here: https://github.com/rdkcentral/Lightning/blob/master/src/platforms/browser/WebPlatform.mjs#L78. When a `.destroy()` is called on Lightning the interval is never cleared. 

This PR clears the interval and makes sure the entire application can be GC'ed.

During my testing with @albertd we ran into other residual Lightning code that was stuck in the heap after calling `destroy()`, likely because the GL Context is created but never released (see: https://github.com/rdkcentral/Lightning/blob/master/src/tree/Stage.mjs#L67). The residual memory is about 400 kb to 1.4 Mb, though negligibly small this PR additionally cleans up as much as we can. During my tests I've seen about 100 Kb Heap increase over +80 start/destroy cycles. Will provide some further testing. 

To validate/test:
1. Run the `/examples/destroy/index.html` test app
2. Create a Heap snapshot
3. Use space to start an app
4. Space again to close the app
5. Create a new Heap snapshot

Compare size of the two snapshots. Additionally force the GC from the console or give the browser a few Mississippi's to run the GC routine. 

Thanks @erikhaandrikman for the additional support and love <3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v2.9.1
+
+*21 apr 2023*
+
+- 🔥 Hotfix for memory leak when `pauseRafLoopOnIdle` is enabled (introduced in v2.7.0)
+- Implemented additional cleanup of Lightning code that gets stuck on the heap after calling `destroy`
+
 ## v2.9.0
 *16 feb 2023*
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@lightningjs/core",
-  "version": "2.9.0",
+  "version": "2.9.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@lightningjs/core",
-      "version": "2.9.0",
+      "version": "2.9.1",
       "license": "Apache-2.0",
       "devDependencies": {
         "@babel/core": "^7.8.3",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "author": "Metrological, Bas van Meurs <b.van.meurs@metrological.com>",
   "name": "@lightningjs/core",
-  "version": "2.9.0",
+  "version": "2.9.1",
   "license": "Apache-2.0",
   "main": "dist/lightning.js",
   "module": "index.js",


michielvandergeest (approved):
Nice findings! 😃 

The issue with residual Lightning code on the heap due to not fully cleaning up the gl context made me think whether the same cleanup issue might occur when using Canvas2D (https://github.com/rdkcentral/Lightning/blob/master/src/tree/Stage.mjs#L65).
Might be good to double check (and possibly fix) that as well.

src/tree/core/CoreContext.mjs
```diff
@@ -49,6 +49,21 @@ export default class CoreContext {
...
+        this.renderExec = null;
+        this._renderTexturePool = null;
+        this._zSorts = null;
+
+        delete this.stage;
+        delete this.root;
+        delete this.renderState;
+        delete this.renderExec;
+        delete this._renderTexturePool;
+        delete this._zSorts;
```
frank-weindel, line 66: Why the `null` set and then a delete?

frank-weindel (approved):

michielvandergeest (approved):

erikhaandrikman (approved):

imtiaz101325:
Hey @uguraslan 👋🏾 
Any change this will get added to a Lighting-SDK release soon?
Our contacts at Metrological have been chasing us about memory leak on the LGI platforms and this might help 🙏🏾

erikhaandrikman:
Core dependency currently is `^2.7.0`: https://github.com/rdkcentral/Lightning-SDK/blob/92d2d9a35809568b8e9e1b3c65529f0b5221858a/package.json#L28 this means it will update to all future minor/patch versions: https://github.com/npm/node-semver#caret-ranges-123-025-004

running `npm update` or `npm update @lightningjs/core` would do the trick. 

Using the `@lightningjs/cli` to create a new project will automatically install https://github.com/rdkcentral/Lightning/releases/tag/2.10.0

imtiaz101325:
Thanks a lot @erikhaandrikman 🙏🏾

PR #495, opened 2023-06-22
Remove unnecessary conditional

itsjamie:
The conditional is provable `if (true)`.

`pos` starts at 0, and only increments in the for loop below bounded in iterations by `length`.

Therefore, pos can never be greater than length.


erikhaandrikman (approved):
Great find! We will ship this with the upcoming release

PR #501, opened 2023-07-03
Better ES export proposal

elsassph:
This PR updates the ES export to expose both a default export with the existing API, and individual classes exports:

## General API

```typescript
import { Application, Component} from "@lightningjs/core";
// or
import Lightning from "@lightningjs/core";
```
Both can be used in conjunction without increasing bundle size, but the default export can not be tree-shaken.

## TS/bundler config

On the consumer project side the following configuration is needed:
```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "moduleResolution": "NodeNext",
  }
}

// package.json
{
  "type": "module"
}
```

VS Code and Vite then nicely pick up the ES exports and offer automatic imports when using e.g. `Component` alone.

Webpack requires adding `.mjs` to the recognised extensions:
```jsonc
// webpack.config.js
...
  resolve: {
    extensions: [".ts", ".js", ".mjs"],
  },
...
```

## Inspector

Simple usage, not tree-shaking friendly:

```typescript
import "@lightningjs/core";
import "@lightningjs/core/devtools";

// note: default export registers `window.lng` which is automatically picked by `attachInspector`
```

Tree-shaking friendly usage:

```typescript
import {
  Application,
  Component,
  Element,
  ElementCore,
  ElementTexturizer,
  Stage,
  Texture
} from "@lightningjs/core";
import "@lightningjs/core/devtools";

// note: Application and Component are optional
window.attachInspector({
  Application,
  Component,
  Element,
  ElementCore,
  ElementTexturizer,
  Stage,
  Texture,
});
```


frank-weindel:
Excellent! Thanks for digging into this, @elsassph. I like this approach. What do you think of calling the export `pure-esm` instead of `es`? I don't think `es` alone quite emphasizes and describes the implications that importing it will have.

Good catch on the Inspector. We might be able to get away with creating a new window global object (with a new name) that only includes the core classes needed by it.

elsassph:
Happy that you like that first shot.

I don't have a strong opinion on `es` but I appreciate that it's a bit ambiguous and potentially dangerous if you mistakenly import both the global `lng` export and an ES export. The proper way would be a separate npm module, like `@lightningjs/core-es`. Like `lodash-es` for `lodash`.

For inspector, it would be optimal to have a "regular export" `attachInspector(/* explicit types references*/)`, then you could add a `lng.attachInspector()` which could be automatic.

frank-weindel:
@elsassph I reopened this one in lieu of #503. I still think we should call the export "pure-esm" to be more clear what this is. The existing export is ESM compliant technically already. This one purifies it to allow tree shaking.

elsassph:
@frank-weindel OK I revisited the PR to clean up the ES export.

- commonjs export (`require`) still works, pointing to `lightning.min.js`
- ES export allows using the `default` export with the complete API-compatible object, and the individual exports.

I verified with a basic "hello world" app, both with Webpack (commonjs and ES versions) and Vite (ES only):

1. minimal ES without application
   ```typescript
   import { Stage, Element } from "@lightningjs/core";
   ```
2 minimal ES
   ```typescript
   import { Application, Component } from "@lightningjs/core";
   ```
3 full ES / commonjs
   ```typescript
   import Lightning from "@lightningjs/core";
   ```

image

frank-weindel:
Thank you! This looks great. I'm not sure if we should be exporting the minified bundle for commonjs tho. Typically bundlers on the user end do minification for the entire application bundle.

elsassph:
Now I not sure now about minification - I've seen it done often but you're probably right: 
https://stackoverflow.com/questions/48673408/should-javascript-npm-packages-be-minified

Reverted that change.

elsassph:
Also added the inspector

devtools/lightning-inspect.js
```diff
@@ -843,18 +843,20 @@ window.attachInspector = function({Application, Element, ElementCore, Stage, Com
...
-        _updateFocus.apply(this, arguments)
-        const focused = this._focusPath && this._focusPath.length ? this._focusPath[this._focusPath.length - 1] : null;
-
-        if (prev != focused) {
-            if (prev) {
-                val(prev, 'focused', false, false);
-            }
-            if (focused) {
-                val(focused, 'focused', true, false);
+    if (typeof Application !== "undefined") {
```
frank-weindel, line 846: What is the purpose of this change? I'd think maybe to avoid errors of Application is not defined, but the same errors can occur if the other classes are not defined.

frank-weindel: Never mind I see! Could you add a comment explaining why Application is optional?

elsassph: Well, `Component` was already optional, and we use neither `Component` nor `Application` - the other objects are core to the rendering so there's no reason to make them optional.

Issue #490, opened 2023-06-13
Tree shaking


philippe-wm:
Currently Lightning core is published to npm as a monolithic JS distribution where everything is exported as one object, which doesn't allow tree-shaking of the unused functionality.

We currently optimise for canvas 2D compatible devices and by doing a custom build of Lightning removing the unused features we can reduce this dependency payload by 30% at least.

We appreciate that a simple pre-minified JS is good for many, but it would be nice to support tree-shaking without doing a custom Lightning build/fork.

frank-weindel:
@philippe-wm Please give the linked PR branch a try and let me know if it works for you.

You should only need to build and pack the project and then install it as a file:
- `npm run build`
- `npm pack` should create  a file `lightningjs-core-2.10.0.tgz`
- Move the file to your project directory and run: `npm install file:lightningjs-core-2.10.0.tgz`

elsassph:
@frank-weindel what branch are you referring too?

PR #503, opened 2023-07-04
Introducing `core-es` package

elsassph:
Following discussions on https://github.com/rdkcentral/Lightning/issues/501 with @frank-weindel, in order to address https://github.com/rdkcentral/Lightning/issues/490, I believe we should have 2 explicit libraries:

- `@lightningjs/core` - the current library
- `@lightningjs/core-es` - a pure ES library, optimal for tree-shaking

PR has 2 commits:

1. Reorganises project into "packages", with current `core` library source moved under `packages/core`
2. Adds `package/core-es`

The idea is that from this one repo, the 2 packages could be published.


elsassph:
Is it maybe too radical?

frank-weindel:
Thanks for submitting this @elsassph. I discussed with the team and we'd like to go with the simpler approach you laid out in #501.

elsassph:
Thanks @frank-weindel. What about a separate repo for the ES exports?

There is a real concern IMHO that someone mistakenly imports both the global `lng` object and the ES source (VS Code can easily import it automatically) - which would unexpectedly significantly increase the size of a project.

frank-weindel:
We would still transition to using the full source tree (built to `dest`) for the original ESM export. The new pure ESM export would import from that same source tree. So the worst that should happen is that users who mix will get all of Lightning in their bundle, as if they had just used the original ESM export anyway.

I'm not sure how much a separate repo would help. In any case, if a user installs it (assuming it's dependent on `@lightningjs/core`) and then installs an existing component library that still uses the `@lightningjs/core`'s original ESM export they will be in the same boat as above.

As you've been able to utilize tree shaking in past Lightning versions without a separate package, what in your opinion would be the best way for you to achieve that again? I'm assuming your project does not use the Lightning SDK nor the two component libraries we have published in RDK Central as they use the original `Lightning`/`'lng` ESM export.

Issue #496, opened 2023-06-27
Enable to use webgl2 when webgl is not enabled in device.


strokirus:
Hello, I've been working in a device with webgl (1) disabled by default but with webgl2 enabled, so I just create a patch to check with webgl2 is enabled in the device and create a context based on it.

I updated the file core/src/platforms/browser/WebPlatform.mjs in the createWebGLContext function line 193.

`let gl = canvas.getContext('webgl2', opts) || canvas.getContext('webgl', opts) || canvas.getContext('experimental-webgl', opts);
`

I hope this could help our mates, if necessary please inform me to create PR.

uguraslan:
Hey @strokirus, thanks for bringing up this issue and providing your patch to handle WebGL compatibility in devices that only support WebGL2.

Regarding the suggested change, while WebGL2 is considered as almost completely backward compatible with WebGL1, it would be still better to place the `webgl2` context option as the last one to minimize the impact on existing functionality.

And please feel free to open a pull request with your proposed change so that we can properly review it and include it in the next release.

In addition to that, if possible, we would appreciate if you could provide more insights into why a specific device only supports WebGL2 or why WebGL1 is disabled by default. Understanding such configurations can help us gather valuable information for future improvements.

strokirus:
Hello @uguraslan and thank you for the reply.

So, I've been testing a Lightning app in an AndroidTV and receive the message 'This browser does not support webGL.', but I notice we have multiples apps already installed with webgl ready, after some investigation I notice the webgl1 is installed but disabled by default I don't know if this was a specification directly from manufacturer but I notice we have webgl2 enabled as a prior choice. After this I create a patch and got a working result.

PR #504, opened 2023-07-07
Adding support to create a context for webgl2.

strokirus:
I'm testing the Lightning app in a specific MVPD directly on AndroidTV STB and I notice in this specific box the only webgl enabled was webgl2. So I added this specific context when webgl is created.

Resolve https://github.com/rdkcentral/Lightning/issues/496


uguraslan (approved):
LGTM

PR #508, opened 2023-07-21
Fix texture changed logic

elsassph:
It looks like there has been (forever) a bug in the elements' `txLoaded` event, over-firing because it incorrectly identifies texture changes.

`this.__core.displayedTextureSource` is a `TextureSource` but was apparently used in the code as if it was a `Texture` (thank you untyped JS).

### DIFF
--- a/src/tree/Element.mjs
+++ b/src/tree/Element.mjs
@@ -606,7 +606,7 @@ export default class Element {
             }
         }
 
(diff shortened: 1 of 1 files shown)


michielvandergeest (approved):
LGTM!

@erikhaandrikman could you please take a look at this one as well so we can include it in the October release

erikhaandrikman (approved):

PR #516, opened 2023-11-21
Fix example imports

wouterlucas:
Recent change https://github.com/rdkcentral/Lightning/commit/e52364ab423c066851d0003863c94e8c7340f966 introduced /dist/src to enable tree shaking.

However since some of the files get transpites from mts the examples can no longer import directly from /src/lightning.mjs. This commit changes all examples to pull from the ./dist/src/ folder instead.


uguraslan (approved):
LGTM

PR #522, opened 2024-01-11
fix: RoundedRectangleShader clipping stroke

ricardo-a-alves-alb:
At it's current state, when setting a value for stroke on the RoundedRectangleShader the stroke itself gets clipped incorrectly.
This is more noticiable when you put the radius at half the height (on a square) to make it a circle and then you can see it clipping the top/bottom and left/right of the stroke.

Comparative

Round:
Screenshot 2024-01-11 at 21 16 40
vs
image

Rectangle:
image
vs
image


frank-weindel (approved):
LGTM

chiefcll (approved):

PR #526, opened 2024-02-23
feature: TextTexture handle right to left text direction

Zacchari-Carde-Wiztivi:
Hi ! in our project we have Arabic texts which are displayed correctly by the canvas but the bidirectional texts always have a LTR global context so text blocs are not displayed well when our UI is in RTL.
This change set the canvas context direction to allow the browser to handle properly bidirectional texts.


wouterlucas (approved):
👍

frank-weindel (approved):
LGTM

PR #528, opened 2024-02-29
Fixes issue/527

adiffmike:
Version 2.12.1 not including typescript types

Switched extension search in scripts/src-to-dist.cjs to use ".ts" and ".mts" so the types get added to the build.


wouterlucas (approved):
Nice! Thanks for the PR 👍

PR #553, opened 2024-10-17
fix(Element): use similar types for displayed texture source comparison

ImCoolNowRight:
The change introduced in #508 seems to be comparing incompatible types, which is causing an issue where patching in a new texture source is the only thing that triggers a `txLoaded` event to fire; patching new dimensions does not fire that event.

The issue can be seen here: https://playground.lightningjs.io/#Mhe2FycwzpSt

This fix reverts the check to a previous version where patching in dimensions constitutes as a new texture having loaded, which will fire off the appropriate event.

### DIFF
--- a/src/tree/Element.mjs
+++ b/src/tree/Element.mjs
@@ -606,7 +606,7 @@ export default class Element {
             }
         }
 
-        const prevSource = this.__core.displayedTextureSource;
+        const prevSource = this.__core.displayedTextureSource ? this.__core.displayedTextureSource._source : null;
         const sourceChanged = (v ? v._source : null) !== prevSource;
 
         this.__displayedTexture = v;


wouterlucas:
I think the current behavior is correct, if you look at the docs: https://lightningjs.io/docs/#/lightning-core-reference/Templates/Events?id=events

```
txLoaded | Fires when a texture is successfully loaded.
```

That means `txLoaded` only should fire if a texture is loaded right? Changing the dimensions doesn't update the texture so nothing is (re)loaded thus doesn't provide a `txLoaded` event.

Might warrant another event, but what's the use case for listening for dimension changes on an Image tag?

ImCoolNowRight:
One example is that we have an Artwork component -- which handles rendering multiple textures, fallback sources, and effects -- that will scale down an image to a user-defined width and height before checking for and applying various effects, but we need to wait until the texture is fully loaded before we can grab the aspect ratio and scale down that texture.

Previously, we were able to wait until the `txLoaded` event was fired before checking the values, but if that should only be fired when the texture source changes another event that would signal that the texture has been manipulated in some way would certainly be a viable alternative!

elsassph:
@ImCoolNowRight you should use `onAfterUpdate` to react to size changes:
https://playground.lightningjs.io/#Vc36DoKlEVFY

PR #562, opened 2025-02-17
Fix types regression

elsassph:
Previous types changes introduced a regression in the "types tests".

This PR fixes some of the TSD tests, but one still eludes me so for now I suggest to disable the TSD tests in CI.


michielvandergeest (approved):
LGTM

PR #563, opened 2025-02-17
Feature - animation by ref

elsassph:
As was mentioned in the TS comments, using a String as animation target can't be type-safe.

It is however useful to be able to animate multiple elements as one single animation, so we have modified the selection logic to allow passing a reference directly for the animation.
```javascript
// before (untyped)
actions: [
  { t: 'SomeUnsafeTarget', ...}
]
// after
actions: [
  { t: this.tag('SomeTypeSafeTarget'), ...}
]
```


wouterlucas (approved):
Nice :) LGTM

PR #564, opened 2025-02-27
Fix: ImageWorker can't load relative URLs

elsassph:
> Relative URLs in Web Workers
> Web Workers run in a separate thread from the main JavaScript execution context. Because of this, they do not have direct access to the DOM, including the BASE tag or the document's URL. This affects how relative URLs are resolved.

Failure to provide a full URL will cause a global `SyntaxError` error to be raised, which won't result in a texture error.


michielvandergeest (approved):
LGTM!

PR #566, opened 2025-03-11
Image URL validation fix for image worker

uguraslan:
In some apps, image URLs are generated like this:

```js
export const getImgUrl = (imgPath, width = 185)=> {
    return `//image.example.org/t/p/w${width}${imgPath}`
};
```

So we should not expect `://` to be present in all image URLs.

### DIFF
--- a/src/platforms/browser/WebPlatform.mjs
+++ b/src/platforms/browser/WebPlatform.mjs
@@ -290,10 +290,16 @@ export default class WebPlatform {
             }
         } else if (this._imageWorker) {
             // WPE-specific image parser.
-            // image urls can be like "//example.com/image.png"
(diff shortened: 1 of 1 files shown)


src/platforms/browser/WebPlatform.mjs
```diff
@@ -290,7 +290,8 @@ export default class WebPlatform {
             }
         } else if (this._imageWorker) {
             // WPE-specific image parser.
-            if (typeof src !== 'string' || src.indexOf('://') < 0) {
+            // image urls can be like "//example.com/image.png"
+            if (typeof src !== 'string' || (src.indexOf('://') < 0 && !src.startsWith('//')))
```
sandeep-vedam: Starting brace missing

sandeep-vedam (approved):
LGTM

michielvandergeest (approved):
LGTM!

PR #570, opened 2025-04-24
Fix text texture RTL propagation

elsassph:
`Element`'s `texture` instance doesn't receive the RTL flag in some scenarios due to the "side-effects optimisation" in `ElementCore`, this causes text alignment to not be automatically mirrored (e.g. `left` alignment becoming `right`).

Note:

- the bug doesn't happen with the classic component templating approach, only when manually creating elements,
- test file was reformatted, the main addition is the new `it("Should propagate flag after attachment"` new last case, plus a few assertions that labels were rendered with the right direction flag.

In this screenshot, before the bug, the "Dynamic attachment" label would remain left-aligned instead of being right aligned with the blue box as we can see here:
image

### DIFF
--- a/src/tree/core/ElementCore.mjs
+++ b/src/tree/core/ElementCore.mjs
@@ -1339,9 +1339,9 @@ export default class ElementCore {
         // update element when the direction changes
         if (!initial) {
             this._triggerRecalcTranslate();
-            // allow side effect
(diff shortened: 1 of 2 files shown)


amurillo-wm:
Tested locally and `text` is mirrored properly

michielvandergeest (approved):
LGTM!

PR #580, opened 2025-07-10
Fix: ZWSP regex bug fix

elsassph:
Why Javascript? Why?
```typescript
const reZWSP = /[\u200B]/;
console.log(1, reZWSP.test('\u200B')); // 1, true
console.log(2, reZWSP.test('\u200B')); // 2, true

const reZWSPGlobal = /[\u200B]/g;
console.log(1, reZWSPGlobal.test('\u200B')); // 1, true
console.log(2, reZWSPGlobal.test('\u200B')); // 2, false
```


michielvandergeest (approved):
LGTM

PR #584, opened 2025-08-20
Feature: don't render very transparent elements

elsassph:
## Context

Often we want an element to be invisible, yet we want its texture to load / remain loaded.

A pattern we use heavily is to make an element (or a tree of elements) to a very low opacity (like `0.001`) to work around the activation logic. This ensures all the textures are loaded and remain in memory, so we avoid the texture loading delay when we need to show the element.

The issue with this pattern is that Lightning will effectively paint those invisible elements, which means they impact the rendering performance.

## Solution

This PR adjusts the quad collection logic (and the mouse collection logic) to ignore elements at a low opacity.
The value `0.002` was chosen to avoid the usual JS floating bugs if one choses to animate an element to `0.001` alpha.

## Alternatives considered

For the use case described, we could imagine introducing an Element flag to indicate that we want the texture to load/remain loaded even when it is invisible. 
- This would in fact represent a lot of extra complexity and changes in sensitive code paths,
- It would be a burden for users, which would have to annotate every element in a sub tree which should be loaded.

To overcome the user-burden we might imagine that the flag could be inherited, but this would mean even more complexity and risks to introduce regressions.

### DIFF
--- a/docs/RenderEngine/Elements/Rendering.md
+++ b/docs/RenderEngine/Elements/Rendering.md
@@ -33,6 +33,18 @@ You can set the visibility of an element in the following ways:
 * Using the `visible` property. If  the value of this property is set to 'false',  the element is not rendered (which saves performance). If an element is invisible, the off-screen elements are invisible as well, so you do not have to hide those manually to maintain a good performance.
 * Using the `alpha` property, which defines the opacity of an element and its descendants. If the value of this property is set to 0 (zero), the element is not rendered.
 
+### Alpha-based Performance Optimization
+
+Lightning automatically optimizes rendering performance by skipping elements with very low alpha values. Elements with an effective alpha of `0.001` or less are considered invisible and are automatically excluded from:
+
+* Rendering operations
+* Mouse hit testing and interaction detection
+
+This optimization helps improve performance by avoiding unnecessary rendering work for elements that are effectively invisible to users, while still allowing for texture loading and smooth fade-in/fade-out animations.
+
+Note: actual alpha test is `< 0.002` to avoid floating point errors when animating to `0.001`.
+
+
 ## Color
 
 


michielvandergeest:
@elsassph quick question before merging this in: does this PR also affect the loading of images on elements with a low alpha? Or it does still trigger an image fetch operation on elements with a `src` attribute and an `alpha` of  `0.0001` ?

elsassph:
@michielvandergeest elements/images are still considered visible/active at a very low alpha. The optimisation avoids rendering something invisible.

michielvandergeest:
Ok, so using the "very low alpha to pre-load an image"-trick will still work then

elsassph:
@michielvandergeest that's the whole point

michielvandergeest:
yeah, just wanted to double check 😉

girouxc:
hey @wouterlucas! do you think it would be interesting to add this feature to lightning 3 as well ? 👀

wouterlucas:
> hey @wouterlucas! do you think it would be interesting to add this feature to lightning 3 as well ? 👀

yeah that's something we can consider, though in L3 I'd rather solve this with a proper flag/API then using an low opacity value.

@jfboeve @michielvandergeest lets think about this for the L3 side. Something like a preload flag or "hide" flag to omit it from rendering but process all the other stuff Blits activation + Rendering basics (2D matrix/position/renderability/etc).

michielvandergeest:
yeah agree - I see I mentioned in a previous comment that we should see if this is something to apply to L3 as well. Just don't think we've gotten around really looking into that.

Blits already has a `show` attribute, that currently maps directly alpha. We could introduce a `show` prop on the render as well and map to that, or we can follow the same approach and play with low alphas. That should be easy to wire up in Blits and make the show attribute into 'hidden', but still 'preloadable'

philippe-wm:
We introduced a concept of "permanent" image, which means they never unload (until they are detached). 

One important subtlety though is that the low alpha affects an entire subtree, and if something is visible false or alpha 0 it should still be unloaded.

wouterlucas:
the L3 renderer already has the ability to mark a texture for prevent cleanup, so they won't be unloaded. Typically used with Font maps and some other core components.

and yeah good point, I think show should include the entire subtree. Alpha obviously does that for free, so we might be introducing something that just adds another check of the same. I just don't like the implicitness of it.

PR #587, opened 2025-09-05
Fix ES5 output by using only Babel in this case

### LINKED ISSUE #536: Lightning Core > 2.11.0 Unsupported on older browsers
Hi,
while updating the lng core and sdk to the latest versions (2.13.1 and 5.5.3) We noticed that Lightning crashes on older browsers. We would like to support Samsung 2017 tvs and LG 2017 tvs. These use a chromium browser v47 and v38 respectively. When using lng dev, there seems to be error in the startApp.js file, where:
```
var url = new URL(document.location.href);
resolve(url.searchParams.has('texture'));
```
url.searchParams is supported from chrome 51. 
image
When changing this to resolve true (for testing), we get a different error where bundle is undefined.
image

This seems to be the case even when using core v2.12.0. This did work on core v2.11.0 (currently in use). This would prevent us from updating our lng core. Which is not ideal.
Can these older browser stay supported? Thanks in advance.

Kind regards
Sem

elsassph:
Fixes #536 - Lightning Core > 2.11.0 Unsupported on older browsers

ESbuild/Rollup doesn't seem to allow Babel do the transpilation to older browsers. This is a regression which I didn't try exactly narrowing.

This PR forces all the JS transformation to be done through Babel - it still is properly minified and the JS size is about the same size (~432Kb for `lightning.es5.min.js`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@babel/plugin-transform-parameters": "^7.8.3",
         "@babel/plugin-transform-spread": "^7.8.3",
         "@babel/preset-env": "^7.8.3",
+        "@babel/preset-typescript": "^7.27.1",
         "@playwright/test": "^1.50.1",
         "@rollup/plugin-babel": "^6.0.3",
         "@types/node": "^20.0.0",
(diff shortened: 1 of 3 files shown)


SemHelevenAndrome:
Hi @elsassph Nice! That seems to work. Also tested in full application.

michielvandergeest (approved):
LGTM

PR #588, opened 2025-09-05
Fix/image data url 575

### LINKED ISSUE #575: Bug: base64 images regression
### Problem/Opportunity

Since we added URL validation for the Image Worker, base64 images were missed and they are considered invalid.

### Steps to reproduce

Set `src` to a base64 image URL.

### Expected Behavior

Image should load

### Actual Behavior

Image fails to load when image worker is enabled

### Notes (Optional)

_No response_

omersaleem:
Fixes #575 

A change that was added to check Image src urls caused a regression when using image data urls - this has broken our app as we use this feature to display QR codes generated on the fly.

This fix preserves the original check but ignores these checks for data urls (that begin with `data:`.

This has broken our My Sky app and we have had to revert to an older version of Lightning core until this is released.

### DIFF
--- a/src/platforms/browser/WebPlatform.mjs
+++ b/src/platforms/browser/WebPlatform.mjs
@@ -295,8 +295,9 @@ export default class WebPlatform {
             }
             
             // URL can start with http://, https://, and //
+            // (Ignore data URLs that start with data:)
             const separatorPos = src.indexOf('//');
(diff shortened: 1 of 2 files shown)


michielvandergeest (approved):
Thank @omersaleem! LGTM

elsassph:
Thanks, could that be cherry picked in the RTL branch?

PR #612, opened 2026-07-30
Fix `freeTextureSource()` leaving stale entries in `textureSourceHashmap`

guilhermesimoes:
When `gc()` is called, it calls `freeUnusedTextureSources`, which at the end calls `_cleanupLookupMap`, cleaning up all unused texture sources from `textureSourceHashmap`.

However, when `freeTextureSource` is called (either when a texture source is replaced or when someone manually calls `texture.free()`), the stale texture source is kept inside `textureSourceHashmap`, leading to a memory leak.

This issue was detected because we use a single text texture for subtitles, and keep changing the text of that texture as playback progresses.

### DIFF
--- a/src/tree/TextureManager.mjs
+++ b/src/tree/TextureManager.mjs
@@ -53,7 +53,7 @@ export default class TextureManager {
         for (let i = 0, n = this._uploadedTextureSources.length; i < n; i++) {
             this._nativeFreeTextureSource(this._uploadedTextureSources[i]);
         }
-        
+
         this.textureSourceHashmap.clear();
         this._usedMemory = 0;
     }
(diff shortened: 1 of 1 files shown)


guilhermesimoes:
I have read the CLA Document and I hereby sign the CLA

src/tree/TextureManager.mjs
```diff
@@ -195,6 +195,10 @@ export default class TextureManager {
             this._nativeFreeTextureSource(textureSource);
         }
 
+        if (textureSource.lookupId) {
```
guilhermesimoes: Not sure if we need this check here, maybe we can remove it

jfboeve (approved):

PR #614, opened 2026-08-11
Fix zIndex ordering for alpha zero elements

eXecutable:
forceZIndexContext=true on Element creates a z-context

new Element is added → enableZSort → _zSort=true

During update(), updateTreeOrder() skips alpha=0 children — their _updateTreeOrder stays stale

During render(), sortZIndexedChildren() sorts using mix of fresh and stale values → wrong order → _zSort=false

Later, textLayer becomes visible → gets updated _updateTreeOrder → but _zSort remains false → _zIndexedChildren array is never re-sorted → paints are in the wrong order


wouterlucas (approved):
LGTM @jfboeve - can we run this next week?

